PR #33, opened 2016-05-18
Added timestamp in actions_history

mcarpentierCoveo:
Also,  I updated HistoryElement to match the specifications.

### DIFF
--- a/src/history.ts
+++ b/src/history.ts
@@ -41,11 +41,13 @@ export class HistoryStore {
     }
 }
 
-export type HistoryElement =  HistoryViewElement | any;
+export interface HistoryElement {
+    name: string;
+    value: string;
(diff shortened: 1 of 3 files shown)


src/history.ts
```diff
@@ -41,11 +41,13 @@ export class HistoryStore {
     }
 }
 
-export type HistoryElement =  HistoryViewElement | any;
+export interface HistoryElement {
+    name: string;
+    value: string;
+    time: number;
```
pastjean: Can you make the time a Date type or String (iso8601), not a timestamp... (i've seen too much errors with epoch,timestamps)

pastjean:
All good! passes my test, merge your thing and ping me , I'll release a new version

PR #59, opened 2016-07-07
Increase coverage - add objectassign coverage

pastjean:


francoislg:
Add "empty object should return empty" ?

const result = assing({}, {}, {});

t.deepEqual(result, {});

test/objectassign_test.ts
```diff
@@ -0,0 +1,38 @@
...
+});
+
+test('native: nested properties should be erased', t => {
+    const srcs = [{'1': 2}, {'1': 3}];
+    const result = assign({}, srcs[0], srcs[1]);
+
+    t.deepEqual(result, {'1': 3});
+});
+
+test('native: multiple object copy', t => {
```
francoislg: Missing expectations (i.e. should copy all properties)

PR #83, opened 2016-09-15
Added way to hook on the module Load

francoislg:
Since the script is fetched in an async way, the coveoua script is sometimes not loaded BEFORE the Recommendations component is initialized, thus giving no recommendation on a semi-random basis.

Hooking on this callback would solve the problem very easily 😉

### DIFF
--- a/src/simpleanalytics.ts
+++ b/src/simpleanalytics.ts
@@ -54,6 +54,14 @@ export class SimpleAPI {
                 throw new Error(`Event type: '${event}' not implemented`);
         }
     }
+
+    onLoad(callback: Function) {
+        if (typeof callback == 'undefined') {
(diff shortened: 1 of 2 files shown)


francoislg:
@pastjean If you approve this change, could you also stamp the next version with this update when merged? 😄

francoislg:
Dans le fond, ça éxécute le callback quand il est rendu à dépiler ses events. J'ai décidé de pas mettre de restriction avant ou après le init, parce que ça change pas grand chose au final 😄 

Sample usage: 

```
coveoua("init", whatever)
coveoua("onLoad", function() { initializeDependentStuffHere(); });
```

francoislg:
@pastjean Ready to merge ? :)

PR #108, opened 2016-10-27
Improve store for page view events

olamothe:
Before this PR, the "page view" event was added to the history store only when the script is first loaded in the page.

This works fine for full page reload, but in the context of a single page app (eg: an angular website), the store is not properly updated when a new page is effectively loaded.

This PR moves this logic to the analytics class, since angular user will need to call the "sendViewEvent" somewhere in their app (eg: inside their angular router).

This ensure that the store is properly updated whenever a page view event is sent.

I also added a single definition file for the library, so it's easier to use for external typescript project(s). This could also be submitted to the DefinitelyTyped repo in the future for easier integrations.

The definition file is also compiled on each build, to ensure that it is at least valid typescript.


src/analytics.ts
```diff
@@ -68,6 +70,19 @@ export class Client implements AnalyticsClient {
 
     sendViewEvent(request: ViewEventRequest): Promise<ViewEventResponse> {
         if (request.referrer === '') { delete request.referrer; }
+
+        // Check if we are in a browser env
+        if (hasDocumentLocation()) {
```
francoislg: This change will log multiple times the same `document.location` if we call multiple `sendViewEvent` from the same page, we should probably handle this case gracefully.

src/library.d.ts
```diff
@@ -0,0 +1,156 @@
...
+        private client;
+        init(token: string | Client, endpoint: string): void;
+        send(event: EventType, customData: any): void;
+        onLoad(callback: Function): void;
+    }
+
+    type EventType = 'pageview';
+}
+
+declare module 'coveo.analytics' {
```
francoislg: Just throwing this idea here:

It might be misleading to offer the `.analytics` and `.history` exports. The default code snippet is loaded asynchronously, meaning these variables _might_ not be defined if the coveo.analytics module is loaded after the script relying on this definition file. Thus, TypeScript files relying on this will build correctly, but will throw undefineds at runtime.

Either:

* Load it synchronously
    * We don't want this, since the script is static and we don't want to slow the whole site to log page views
* Use the call I implemented to fix this problem, `coveoua("onInit", callback);`

olamothe: It's a declaration file, this produce no code when we build it.

It's only so when we do : 

   import {history} from 'coveo.analytics';

or 

    import {analytics} from 'coveo.analytics'

From a typescript project, we have proper typing definitions

francoislg: Alright then :) :+1:

src/analytics.ts
olamothe: You mean someone doing an implementation error and logging the same view event multiple time ?

How do you propose we fix that ? 

I guess we could remove duplicate view event (same url) inside the HistoryStore, so that we do not keep 2 duplicate values in a row. 
This would be slightly wrong if someone reload a single page multiple time, but acceptable I think (since for reveal, it should really be just a single page view after all).

Or if someone leave a "session" on a given page view, then comes back a day later on the same page view (from google let's say), then we would wrongly remove the view event.

olamothe: I think we need to implement a mechanism in the HistoryStore to protect against duplicate values added in a short amount if time (say ... 1 min ?)

olamothe:
I added something to prevent consecutive duplicate values in the history store in a short amount of time (1 min). I also asked spaquet if the reveal service already handle those errors (duplicates), and he told me yes.

Still, I think it's cleaner to try to enforce it with the coveo.analytics lib also.

PR #109, opened 2016-11-01
Change compilation target to es5

olamothe:
Otherwise, importing the lib and using it inside a browser which do not have perfect support for es2015/es6 will break (looking at you Safari).


pastjean:
Im suprised you had no problems with promise compilation :/ i had es6 only for that

PR #146, opened 2017-01-04
Reduce cookie size.

olamothe:
Reduce the history size to 10 max.
Keep less data for URL
Trim values to 75 max char.

See : https://coveord.atlassian.net/browse/UA-1938

### DIFF
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -74,10 +74,9 @@ export class Client implements AnalyticsClient {
         // Check if we are in a browser env
         if (hasDocumentLocation()) {
             const store = new HistoryStore();
-            const location = document.location;
             const historyElement = {
                 name: 'PageView',
-                value: location.host + location.pathname,
+                value: document.location.toString(),
                 time: JSON.stringify(new Date()),
                 title: document.title
             };
--- a/src/analytics.ts
+++ b/src/analytics.ts
@@ -74,9 +74,10 @@ export class Client implements AnalyticsClient {
         // Check if we are in a browser env
         if (hasDocumentLocation()) {
             const store = new HistoryStore();
+            const location = document.location;
             const historyElement = {
                 name: 'PageView',
-                value: document.location.toString(),
+                value: location.host + location.pathname,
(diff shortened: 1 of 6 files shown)


src/analytics.ts
```diff
@@ -74,9 +74,10 @@ export class Client implements AnalyticsClient {
         // Check if we are in a browser env
         if (hasDocumentLocation()) {
             const store = new HistoryStore();
+            const location = document.location;
             const historyElement = {
                 name: 'PageView',
-                value: document.location.toString(),
+                value: location.host + location.pathname,
```
olamothe: This makes it so that page view becomes shorter : 

`https://help.salesforce.com/search#q=test&f:@sflanguage=%5Ben_US%5D&firstQueryMeta=%5Bobject%2520Object%5D` will become `help.salesforce.com/search`

src/history.ts
```diff
@@ -12,6 +13,9 @@ export class HistoryStore {
 
     addElement(elem: HistoryElement) {
         elem.internalTime = new Date().getTime();
+        if (elem.value != null) {
+            elem.value = elem.value.slice(0, MAX_VALUE_SIZE);
```
olamothe: Any value that gets over 75 char will be trimmed (page view or search)

src/history.ts
```diff
@@ -1,8 +1,9 @@
 import {WebStorage, getAvailableStorage} from './storage';
 
 export const STORE_KEY: string = '__coveo.analytics.history';
-export const MAX_NUMBER_OF_HISTORY_ELEMENTS: number = 20;
+export const MAX_NUMBER_OF_HISTORY_ELEMENTS: number = 10;
```
olamothe: Reduce to 10 history elements maximum

jrochette:
Looks good. Travis checks are failing (@pastjean is this expected?) though

src/analytics.ts
paquetse: With what I see in client's data. This fix is too agressive. We need the Location.search part. @olamothe I will contact you in private.

PR #155, opened 2017-08-29
Put back internalTime because it is used internally by the historySto…

paquetse:
Put back internalTime because it is used internally by the historyStore. It should not have been removed completely in the last pull request (153).
Internally, the internalTime is kept, but by defaut, getHistory removes the internalTime, because it is not necessary in the actionHistory sent to Coveo ML.

### DIFF
--- a/src/history.ts
+++ b/src/history.ts
@@ -20,7 +20,7 @@ export class HistoryStore {
     addElement(elem: HistoryElement) {
         elem.internalTime = new Date().getTime();
         this.cropQueryElement(elem);
-        let currentHistory = this.getHistory(false);
+        let currentHistory = this.getHistoryWithInternalTime();
         if (currentHistory != null) {
             if (this.isValidEntry(elem)) {
                 this.setHistory([elem].concat(currentHistory));
@@ -30,21 +30,19 @@ export class HistoryStore {
         }
     }
 
-    getHistory(stripInternalTime: boolean = true): HistoryElement[] {
-        let history: HistoryElement[] = [];
+    getHistory(): HistoryElement[] {
+        let history = this.getHistoryWithInternalTime();
+        return this.stripInternalTime(history);
+    }
+
+    private getHistoryWithInternalTime(): HistoryElement[] {
         try {
-            history = <HistoryElement[]> JSON.parse(this.store.getItem(STORE_KEY));
+            return <HistoryElement[]> JSON.parse(this.store.getItem(STORE_KEY));
         } catch (e) {
             // When using the Storage APIs (localStorage/sessionStorage)
             // Safari says that those APIs are available but throws when making
             // a call to them.
             return [];
-        }
-
-        if (stripInternalTime) {
-            return this.stripInternalTime(history);
-        } else {
-            return history;
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/history.ts
```diff
@@ -18,21 +18,26 @@ export class HistoryStore {
...
                 this.setHistory([elem].concat(currentHistory));
             }
-
         } else {
             this.setHistory([elem]);
         }
     }
 
-    getHistory(): HistoryElement[] {
+    getHistory(stripInternalTime: boolean = true): HistoryElement[] {
```
francoislg: `this.getHistory()` and `this.getHistoryWithoutInternalTime()` is way clearer than `this.getHistory(true)` and `this.getHistory(false)`

src/history.ts
```diff
@@ -18,21 +18,26 @@ export class HistoryStore {
...
         try {
-            return <HistoryElement[]> JSON.parse(this.store.getItem(STORE_KEY));
+            let history = <HistoryElement[]> JSON.parse(this.store.getItem(STORE_KEY));
+            if (stripInternalTime) {
+                return this.stripInternalTime(history);
```
francoislg: If `this.stripInternalTime()` throws, it will be silently catched by the `catch` down there, then returning `[]` which would has high probability of causing headaches ;)

I would suggest wrapping `this.store.getItem(STORE_KEY)` in the try {} catch {} and leaving the rest out of it.

src/history.ts
```diff
@@ -89,6 +94,14 @@ export class HistoryStore {
         }
         return true;
     }
+
+    private stripInternalTime(history: HistoryElement[]): HistoryElement[] {
+        history.forEach(function(part, index, array) {
+                delete part.internalTime;
+                array[index] = part;
```
francoislg: I don't think you need to reassign it.

src/history.ts
```diff
@@ -89,6 +94,14 @@ export class HistoryStore {
         }
         return true;
     }
+
+    private stripInternalTime(history: HistoryElement[]): HistoryElement[] {
+        history.forEach(function(part, index, array) {
```
francoislg: Use arrow functions `(part, index, array) => {`

test/history_test.ts
```diff
@@ -78,6 +78,28 @@ test('HistoryStore should be able to get the history', t => {
...
+            time: JSON.stringify(new Date()),
+            internalTime: new Date().getTime()
+        });
+    }
+
+    for (let elem of historyElements) {
+        t.true(elem.hasOwnProperty('internalTime'));
+    }
+
+    historyStore['stripInternalTime'](historyElements);
```
francoislg: Not a fan of testing internal methods this way.

Can you test using `getHistory()` directly?

src/history.ts
paquetse: The problem is that the internalTime was returned externally, thus I want to change the default behaviour of this function. With a default parameter, all previous calls will continue to work. 
If I change the name, then all clients using this lib will need to change their code.

src/history.ts
paquetse: Done

src/history.ts
paquetse: Done

src/history.ts
paquetse: You are right

test/history_test.ts
paquetse: I was not able to test it in getHistory, because the mock is not a real store, thus getHistory always returns []. This is why I made a separate method to allow to test it. 
Thus, I thought it would be better to have some hacky code in the test compared to a public method in history. But, I might be wrong.

src/history.ts
francoislg: `this.getHistoryWithInternalTime()` then? 😄

test/history_test.ts
francoislg: ☝️ There is no way to input values in that store? 😱 

alright then

PR #157, opened 2018-02-07
Add Promise polyfill

louis-bompart:
I add a polyfill and shim similar of the ones from JSUI.
This PR is related to JSUI-1996.


tedre191:
No need to add the shims, only the promise check

src/browser.ts
```diff
@@ -4,6 +4,14 @@ import { HistoryStore, HistoryViewElement } from './history';
 
 declare const global: any;
 
+declare var require: any;
+const promise = (window as any)['Promise'];
+if (!(promise instanceof Function)) {
+  require('es6-promise').polyfill();
+}
+import { shim } from './promiseshim';
```
tedre191: remove that

src/promiseshim.ts
```diff
@@ -0,0 +1,54 @@
+export function shim() {
```
tedre191: delete that file

tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "target": "es5",
         "outDir": "dist",
         "declaration": true,
-        "noImplicitAny": true,
+        "noImplicitAny": false,
```
tedre191: why ?

src/browser.ts
```diff
@@ -4,6 +4,14 @@ import { HistoryStore, HistoryViewElement } from './history';
 
 declare const global: any;
 
+declare var require: any;
+const promise = (window as any)['Promise'];
```
tedre191: window['Promise']

tsconfig.json
louis-bompart: Because they were a lot of implicit any in the promiseshim.ts file and to get past it with arrow functions was truly too much of a hassle for me to bother. (And JSUI does it too)

src/browser.ts
louis-bompart: Yeah true, now that I have implicit any allowed.

src/promiseshim.ts
louis-bompart: Done

src/browser.ts
louis-bompart: Now that the noimplictany is enforced, that's needed because otherwise window is implicitly converted as any

tsconfig.json
louis-bompart: Now reverted

src/index.ts
```diff
@@ -3,6 +3,7 @@ import * as SimpleAnalytics from './simpleanalytics';
 import * as history from './history';
 import * as donottrack from './donottrack';
 import * as storage from './storage';
+
```
tedre191: New line ?

src/browser.ts
```diff
@@ -4,6 +4,12 @@ import { HistoryStore, HistoryViewElement } from './history';
...
+if (!(promise instanceof Function)) {
```
tedre191: Add a comment (take the one from search-ui project)

src/index.ts
louis-bompart: Can be removed, but I left it to separate export and import

src/browser.ts
louis-bompart: 👍

PR #159, opened 2018-04-30
Update dependencies for Webpack 4

francoislg:
Last release is broken in Travis because of an update in one TypeScript definition.

See here: https://travis-ci.org/coveo/coveo.analytics.js/builds/372129667?utm_source=github_status&utm_medium=notification

I tried to fix the issue, but figured it would be easier to just update the dependencies of the project.

For reference, here is the definition update that broke our build by using features available only in TypeScript 2.3+ : https://github.com/DefinitelyTyped/DefinitelyTyped/pull/25292/files

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "typescript.tsdk": "node_modules\\typescript\\lib"
+}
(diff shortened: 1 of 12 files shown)


francoislg:
So hum, it's been a while now :) 

@olamothe @paquetse ?

francoislg:
Its been another while now

PR #160, opened 2019-05-07
Support all types of events with `coveoua` & General clean up of the project

francoislg:
So I needed to send any type of event using the simple `coveoua` call and decided that all calls should instead pass through the generic `client.sendEvent` call.

I tried not to be *too* breaking, I changed some stuff that will not impact the two major usage: `history` (used in `coveo-search-ui`) and `coveoua("send", "pageview", { ... })` (used in many projects for page views).

So here is a pretty good list of what was done here: 

* [sendEvent](#diff-612d8d5b8b3b9f8941c32412c5463d2dR58) now handles any logic that was either in the `simpleanalytics` part, or in the specific implementations `sendViewEvent`. 
* **Breaking: ** `sendEvent` now returns the JSON response instead of the "Response" object. Thoughts?
* **Breaking: ** `sendViewEvent` and other implementations do not return the raw `Response` object through an added `raw` attribute. I'd rather have the direct response that matches the response returned from our APIs. Thoughts?
* `language` now always fallbacks on `document.documentElement.lang` instead of only on page views.
* `SimpleAnalytics` renamed to `CoveoUA`. I always found `Simple` to be quite meaningless :)
* Removed `popFromObject` and `objectassign` since they were now useless with new TypeScript features
* Put some files into sub-folders `client` and `coveoua`, to better separate some concerns.
* Added exports for `CoveoAnalyticsClient`, `CoveoUA` and `handleOneAnalyticsEvent` directly on `coveoanalytics` since those are the most logical things to import and were scattered in sub-variables
* Indent on 4 spaces, some files were 2, some were 4, I just put them all to 4.


francoislg:
Had some issues with the travis build, but now it's fixed and ready to be reviewed :)

src/coveoua/simpleanalytics.ts
```diff
@@ -0,0 +1,96 @@
...
+
+    // init initializes a new SimpleAPI client.
+    // @param token is your coveo access_token / api_key / ...
+    // @param endpoint is the endpoint you want to target defaults to the
+    //        usage analytics production endpoint
+    init(token: string | AnalyticsClient, endpoint: string): void {
+        if (typeof token === 'undefined') {
+            throw new Error(`You must pass your token when you call 'init'`);
+        }
+        if (typeof token === 'string') {
```
olamothe: I prefer when a function return/throw early, instead of "somewhere in the middle".

AKA: Move/execute all checks early in the function body, then once you know everything is okay, put the real "stuff to execute" at the end, without any nested if. (I know @samisayegh  will agree on this one :) )

src/coveoua/simpleanalytics.ts
```diff
@@ -0,0 +1,96 @@
...
+            });
+        } else if (typeof token === 'object' && typeof token.sendEvent !== 'undefined') {
+            this.client = token;
+        } else {
+            throw new Error(`You must pass either your token or a valid object when you call 'init'`);
+        }
+    }
+
+    send(event: EventType | DeprecatedEventType, payload: any = {}): Promise<AnyEventResponse> {
+        if (typeof this.client == 'undefined') {
```
olamothe: Same comment as above.

tsconfig.json
```diff
@@ -12,5 +12,8 @@
     },
     "include": [
         "src/**/*.ts"
+    ],
```
olamothe: I would change the tsconfig to strictest mode available. You'll catch lots of bugs/issue with that change.

src/client/analytics.ts
```diff
@@ -0,0 +1,151 @@
...
+    ViewEventRequest,
+    ViewEventResponse,
+    VisitResponse
+    } from '../events';
+import { HistoryStore } from '../history';
+import 'whatwg-fetch';
+
+export const Version = 'v15';
+
+export const Endpoints = {
```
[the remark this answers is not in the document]
francoislg: In fact, I removed them, I find it weird to expose those "internal" endpoints in a script like that, but I could put them back

src/client/analytics.ts
```diff
@@ -0,0 +1,151 @@
...
+
+        const response = await fetch(`${this.getRestEndpoint()}/analytics/${eventType}`, {
+            method: 'POST',
+            headers: this.getHeaders(),
+            mode: 'cors',
+            body: JSON.stringify(body),
+            credentials: 'include'
+        });
+        if (response.ok) {
+            return await response.json() as AnyEventResponse;
```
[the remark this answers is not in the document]
francoislg: My next pull request handles exactly that, using a simple buffer that uses `navigator.sendBeacon()` on unload and a simple fetch in other cases

src/coveoua/browser.ts
```diff
@@ -10,7 +9,7 @@ declare const global: any;
 declare var require: any;
 const promise = (window as any)['Promise'];
 if (!(promise instanceof Function)) {
-  require('es6-promise').polyfill();
+    require('es6-promise').polyfill();
```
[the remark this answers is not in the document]
francoislg: > Screw IE11

👍 

I had another polyfill for `sendBeacon`. With both polyfill, 32kb. Without, 16kb. Totally worth it. I added some error logs instead to include a polyfill if required.

src/coveoua/simpleanalytics.ts
francoislg: Hmm.... I really tried, but in this case, TypeScript is using `typeof token === 'string'` to infer that `token` is a string and not `AnalyticsClient`, so I can't put intermediary variables and re-use them.

Putting the last `throw` first would look like this, which I find harder to read:

`if (typeof token !== 'string' && !(typeof token === 'object' && typeof token.sendEvent !== 'undefined')) {`

or 

`if (typeof token !== 'string' && (typeof token !== 'object' || typeof token.sendEvent === 'undefined')) {`

src/client/analytics.ts
```diff
@@ -0,0 +1,151 @@
...
+    async getHealth(): Promise<HealthResponse> {
+        const response = await fetch(`${this.getRestEndpoint()}/analytics/monitoring/health`);
+        return await response.json();
+    }
+
+    private getBodyForTypeOfEvent(eventType: EventType, request: any) {
+        if (eventType === 'view') {
+            return {
+                location: window.location.toString(),
+                referrer: document.referrer,
```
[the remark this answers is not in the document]
francoislg: `search`, `click` and `custom` all use `originLevel3` according to the documentation. I guess we could put it there, but I find it a little weird to fill it even if you don't provide `originLevel1` and 2 🤔

src/client/analytics.ts
```diff
@@ -0,0 +1,151 @@
...
+    hipaa: 'https://usageanalyticshipaa.coveo.com'
+};
+
+export interface ClientOptions {
+    token?: string;
+    endpoint?: string;
+    version?: string;
+}
+
+export type EventType = 'search' | 'click' | 'custom' | 'view';
```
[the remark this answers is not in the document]
ThibodeauJF: +1 for enum

src/coveoua/simpleanalytics.ts
```diff
@@ -0,0 +1,97 @@
...
+import { AnyEventResponse } from '../events';
+
+/** @deprecated */
+export type DeprecatedEventType = 'pageview';
+
+export type AvailableActions = keyof(CoveoUA);
+
+// CoveoUA mimics the GoogleAnalytics API.
+export class CoveoUA {
+    private client: AnalyticsClient | undefined;
```
ThibodeauJF: I think you could just put 
`private client: AnalyticsClient?;`

src/coveoua/simpleanalytics.ts
ThibodeauJF: Could you export those conditions inside clearly name variables before the if/else

src/coveoua/simpleanalytics.ts
ThibodeauJF: `if(!this.client)` should be sufficient, no? Do you have a difference between an undefined or a null?

src/history.spec.ts
```diff
@@ -93,24 +93,24 @@ test('HistoryStore should be able to remove all internalTime', t => {
...
+    historyStore.clear();
+    storageMock.verify();
 });
 
 test('HistoryStore should be able to set the history', t => {
-  var historyElements: history.HistoryElement[] = [data];
-  storageMock.expects('setItem').once().withArgs(history.STORE_KEY, JSON.stringify(historyElements));
-  historyStore.setHistory(historyElements);
-  storageMock.verify();
+    var historyElements: history.HistoryElement[] = [data];
```
ThibodeauJF: I know here is formatting but if you can, change the var to let/const when you see them

src/coveoua/simpleanalytics.ts
```diff
@@ -0,0 +1,97 @@
+import {
```
samisayegh: Usually in the project we use PascalCase for the file names.

src/coveoua/simpleanalytics.ts
```diff
@@ -0,0 +1,97 @@
...
+
+// CoveoUA mimics the GoogleAnalytics API.
+export class CoveoUA {
+    private client?: AnalyticsClient;
+
+    // init initializes a new SimpleAPI client.
+    // @param token is your coveo access_token / api_key / ...
+    // @param endpoint is the endpoint you want to target defaults to the
+    //        usage analytics production endpoint
+    init(token: string | AnalyticsClient, endpoint: string): void {
```
samisayegh: The OR type on the `token` argument makes this function do too much. If you want to provide two types of initialization, the best would be to offer two separate functions. e.g.

```
interface CoveoUAInitOptions {
    token: string;
    endpoint: string;
}

function init(options: CoveoUAInitOptions) {}

function initWithClient(client: AnalyticsClient) {}
```

francoislg: Backward compatibility stuff :(

The best I could do is splitting the body inside `if (typeof (...))`s into their own private method, but they are not _that_ complicated anyway :/

README.md
```diff
@@ -73,6 +73,32 @@ Initially, the `pageview` events data will be used exclusively by the Coveo Reve
 
   c. Reload the page, in the **Name** panel, ensure that you see a **view** event sent to Coveo analytics.
 
+### Pushing Any Coveo Analytics Event
+
+Add the code snippet to all your website pages.
+
+  ```html
+  <script>
+  (function(c,o,v,e,O,u,a){
```
francoislg: The idea is for it to be as small as possible on the page. It mimics how Google Tag Manager works. TBH I would rather not touch this thing at all :P

src/coveoua/browser.ts
```diff
@@ -1,39 +1,38 @@
...
+    console.error(`This script uses window.fetch which is not supported in your browser. Consider adding a polyfill like "fetch".`);
 }
 
 // CoveoUAGlobal is the interface for the global function which also has a
 // queue `q` of unexecuted parameters
 export interface CoveoUAGlobal {
     (action: string, ...params: string[]): void;
     // CoveoAnalytics.q is the queue of last called actions before lib was included
-    q?: string[][];
+    q?: [string, any[]][];
```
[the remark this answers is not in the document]
francoislg: Adding `strict: true` was failing line 35 since `handleOneAnalyticsEvent` has the following signature: `string, ...any`. 

Since `string` is more precise than `any`, the arguments were considered invalid.

This type here could be read as "an array in which the first element is a string instead of any".

README.md
samisayegh: Could this script use human-readable code so that it is easier to understand and debug?

src/client/analytics.ts
```diff
@@ -0,0 +1,170 @@
...
+    getVisit(): Promise<VisitResponse>;
+    getHealth(): Promise<HealthResponse>;
+}
+
+export class CoveoAnalyticsClient implements AnalyticsClient {
+    private endpoint: string;
+    private token: string;
+    private version: string;
+
+    constructor(opts: ClientOptions) {
```
samisayegh: I believe there is room to simplify the constructor, removing the validation in place and manual assignment of options. I suggest the following:

1- Make all properties in the `ClientOptions` required and set the constructor type to `Partial<ClientOptions>`. Optional properties in an interface make it harder to use inside the source code.
2- Create a defaultOptions object/function. I usually use a getter.
```
private get defaultOptions(): ClientOptions {
    return {
        token: '',
        endpoint: Endpoints.default,
        version: Version
    };
}
```

3- Compose the object passed into the constructor with the default options.
4- Store the result in an `options` property.

Together, this removes the need for the check on line 54, the manual assignment on lines 58-60, and the multiple properties on lines 49-51. If we want to add properties to the interface in the future, we only need to update the `defaultOptions` getter and the interface.

src/client/analytics.ts
```diff
@@ -0,0 +1,170 @@
...
+        if (typeof opts === 'undefined') {
+            throw new Error('You have to pass options to this constructor');
+        }
+
+        this.endpoint = opts.endpoint || Endpoints.default;
+        this.token = opts.token;
+        this.version = opts.version || Version;
+    }
+
+    async sendEvent(eventType: EventType, payload: any): Promise<AnyEventResponse> {
```
samisayegh: Is the `payload` an object? If so, it would be better to use `Record<string, string>` or `Record<string, any>` if the values are not always strings.

src/client/analytics.ts
```diff
@@ -0,0 +1,170 @@
...
+
+        const response = await fetch(`${this.getRestEndpoint()}/analytics/${eventType}`, {
+            method: 'POST',
+            headers: this.getHeaders(),
+            mode: 'cors',
+            body: JSON.stringify(cleanedPayload),
+            credentials: 'include'
+        });
+        if (response.ok) {
+            return await response.json() as AnyEventResponse;
```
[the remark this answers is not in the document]
francoislg: Removing `await` means I need to cast this to `Promise<AnyEventReponse>` instead. 

I kinda just like better `await X as AnyEventResponse` instead of `X as Promise<AnyEventResponse>` :)

src/client/analytics.ts
```diff
@@ -0,0 +1,170 @@
...
+            userAgent: navigator.userAgent
+        };
+        if (eventType === 'view') {
+            return {
+                location: window.location.toString(),
+                referrer: document.referrer,
+                title: document.title,
+                ...baseDefaultValues,
+                ...payload
+            } as ViewEventRequest;
```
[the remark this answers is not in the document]
francoislg: I like having the object auto-completable here, I agree that returning `Record<string, any>` would be more accurate, I'll check this out.

PR #161, opened 2019-05-08
Slightly defer analytics logging to allow using `sendBeacon` on page unload

francoislg:
When you click on a link on the page and try to log an event (like a custom or click event), `fetch` requests get canceled because of redirection.

`navigator.sendBeacon` solves this issue, *but* we lose the ability to get the response of the POST request.

The idea here is to accumulate a tiny buffer for the requests, process them with `fetch`, and flush the remaining ones with `navigator.sendBeacon` if the page is unloading.


src/coveoua/simpleanalytics.ts
```diff
@@ -37,8 +36,8 @@ export class CoveoUA {
         }
     }
 
-    send(event: EventType | DeprecatedEventType, payload: any = {}): Promise<AnyEventResponse> {
-        if (!this.client) {
+    send(event: EventType | DeprecatedEventType, payload: any = {}): Promise<AnyEventResponse | void> {
+        if (typeof this.client == 'undefined') {
```
francoislg: 😱 will revert this

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
+            endpoint,
+            version
+        } = {
+            endpoint: Endpoints.default,
+            version: Version,
+            ...opts
+        };
+
+        this.visitorId = '';
+        this.baseUrl = `${endpoint}/rest/${version}`;
```
samisayegh: I prefer to extract and convert this to a getter instead of storing to a property. This removes any doubts that the value mistakenly gets overwritten somewhere else.

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
+            payload: cleanedPayload,
+            handled: false
         });
-        if (response.ok) {
-            return await response.json() as AnyEventResponse;
-        } else {
-            console.error(`An error has occured when sending the "${eventType}" event.`, response, payload);
-            throw new Error(`An error has occurred when sending the "${eventType}" event. Check the console logs for more details.`);
+
+        await this.deferExecution();
```
samisayegh: Hmm, why is this needed?

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
+        return await this.sendFromBufferWithFetch();
+    }
+
+    private deferExecution(): Promise<void> {
+        return new Promise(resolve => setTimeout(resolve, 0));
+    }
+
+    private flushBufferWithBeacon(): void {
+        while (this.bufferedRequests.length > 0) {
+            const { eventType, payload } = this.bufferedRequests.pop() as BufferedRequest;
```
samisayegh: Is the type cast needed if bufferedRequests is declared of type `BufferedRequest[];`?

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
+            this.analyticsBeaconClient.sendEvent(eventType, payload);
         }
     }
 
-    async sendSearchEvent(request: SearchEventRequest): Promise<SearchEventResponse> {
+    private async sendFromBufferWithFetch(): Promise<AnyEventResponse | void> {
+        if (this.bufferedRequests.length > 0) {
```
samisayegh: It would be a little cleaner to extract the condition as a getter `areBufferedRequestsPending` or something along those lines, and use it here and on line 113.

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
+            const popped = this.bufferedRequests.pop();
+            if (popped) {
```
samisayegh: Is this condition needed if we already know the array is populated on line 120?

src/client/analyticsBeaconClient.spec.ts
```diff
@@ -0,0 +1,20 @@
...
+test('AnalyticsBeaconClient: can send an event', async t => {
+    const sendBeaconSpy = sinon.spy(navigator, 'sendBeacon');
+    const baseUrl = "https://bloup.com";
+    const token = '👛';
+    const currentVisitorId = 'hereiam';
+    const eventType = '🥔';
+
+    const client = new AnalyticsBeaconClient(baseUrl, token, { currentVisitorId });
+
+    await client.sendEvent(eventType as EventType, {
```
samisayegh: Could the cast be removed here? I prefer to avoid them as much as possible, even in test files, because it reduces Typescript's ability to do it's job when the code changes.

src/client/analyticsBeaconClient.ts
```diff
@@ -0,0 +1,28 @@
+import { AnalyticsRequestClient } from './analyticsRequestClient';
+import { EventType } from '../events';
+import { VisitorIdProvider } from './analyticsRequestClient';
+
+export class AnalyticsBeaconClient implements AnalyticsRequestClient {
+    constructor(private baseUrl: string,
```
samisayegh: Since there are multiple arguments, I prefer to use an object here that is more easily extensible.

src/client/analyticsBeaconClient.ts
```diff
@@ -0,0 +1,28 @@
...
+        if (!navigator.sendBeacon) {
+            throw new Error(`navigator.sendBeacon is not supported in this browser. Consider adding a polyfill like "sendbeacon-polyfill".`);
+        }
+        const parsedRequestDataKey = this.getParsedRequestDataKey(eventType);
+        const parsedRequestData = `${parsedRequestDataKey}=${encodeURIComponent(JSON.stringify(payload))}`;
+        const visitorId = this.visitorIdProvider.currentVisitorId;
+        const queryParams = `?access_token=${this.token}${visitorId ? `&visitorId=${visitorId}` : ''}`;
+        const url = `${this.baseUrl}/analytics/${eventType}${queryParams}`;
+        // tslint:disable-next-line: no-console
+        console.log(`Sending beacon for "${eventType}" with: `, JSON.stringify(payload));
```
samisayegh: It would be better to use our project logger `new Logger(this).log(message)` and remove the tslint comment :D

src/client/analyticsFetchClient.ts
```diff
@@ -0,0 +1,53 @@
...
+    version?: string;
+    visitorId?: string;
+}
+
+export interface VisitorIdProvider {
+    currentVisitorId: string;
+}
+
+export class AnalyticsFetchClient implements AnalyticsRequestClient {
+    constructor(private baseUrl: string,
```
samisayegh: Same comment as above on using an object. Maybe the constructor can also be part of the `AnalyticsRequestClient` interface as well? Not sure if you need a separate `AnalyticsRequestClientCtor` interface?

```
interface AnalyticsRequestClientCtor {
    new(options: SomeType): AnalyticsRequestClient;
}
```

src/client/analyticsFetchClient.ts
```diff
@@ -0,0 +1,53 @@
...
+            mode: 'cors',
+            body: JSON.stringify(payload),
+            credentials: 'include'
+        });
+        if (response.ok) {
+            const visit = await response.json() as AnyEventResponse;
+            this.visitorIdProvider.currentVisitorId = visit.visitorId;
+            return visit;
+        } else {
+            console.error(`An error has occured when sending the "${eventType}" event.`, response, payload);
```
samisayegh: Same comment as above on using the project logger, the `payload` type to Record, and whether or not we need the `await` on line 35? Also, could the `protected` on line 44 become `private`?

test/setup-browser-env.js
```diff
@@ -1,5 +1,11 @@
...
+const { window } = new JSDOM('<body></body>');
+global.window = window;
+global.document = window.document;
 global.navigator = window.navigator;
 global.fetch = require('isomorphic-fetch');
-global.location = window.location;
\ No newline at end of file
+global.location = window.location;
+global.Blob = window.Blob;
+global.navigator.sendBeacon = function mock() { return true; };
```
samisayegh: Maybe an arrow function would be good here? `global.navigator.sendBeacon = () => true`

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
+        this.visitorId = '';
+        this.baseUrl = `${endpoint}/rest/${version}`;
+        this.bufferedRequests = [];
+
+        this.analyticsBeaconClient = new AnalyticsBeaconClient(this.baseUrl, token, this);
+        this.analyticsFetchClient = new AnalyticsFetchClient(this.baseUrl, token, this);
+        window.addEventListener('beforeunload', () => this.flushBufferWithBeacon());
+    }
+
+    get currentVisitorId() {
```
ThibodeauJF: since you have a get/set for this.visitorId that doesnt modify the value in any way you could just put `visitorId` public instead

src/client/analytics.ts
```diff
@@ -28,88 +32,139 @@ export interface ClientOptions {
...
-        } else {
-            console.error(`An error has occured when sending the "${eventType}" event.`, response, payload);
-            throw new Error(`An error has occurred when sending the "${eventType}" event. Check the console logs for more details.`);
+
+        await this.deferExecution();
+        return await this.sendFromBufferWithFetch();
+    }
+
+    private deferExecution(): Promise<void> {
+        return new Promise(resolve => setTimeout(resolve, 0));
```
ThibodeauJF: What was the problem here?

francoislg: I don't want to instantly flush the buffer.

The idea is:

1. Buffer the request
2. Wait until everything is executed
3. Send the request using fetch

The trick is that if the click is redirecting to another page, the `unload` handle will be triggered *before* step 2 ends, thus flushing the buffer using `sendBeacon` instead.

src/client/analytics.ts
francoislg: Safe-guard when I introduced `strict` mode, since `pop` is `T | undefined`

src/client/analytics.ts
francoislg: @samisayegh ☝️ I think this answers your question about `deferExecution`

src/client/analyticsFetchClient.ts
francoislg: I prefer not to enforce any kind of constructor for interfaces unless there is some generic way to construct them dynamically.

In this case, the implementation could be deeply different between the two clients, they just happen to take the same parameters. Think of a third "NullClient" that would take no parameter and just implement `sendEvent`.

src/client/analytics.ts
samisayegh: Hmm, in that case is the outer condition not neeeded?

src/client/analytics.ts
samisayegh: I see, thanks.

src/client/analyticsBeaconClient.ts
francoislg: I can't directly use the Logger since this is a standalone script :P

src/client/analytics.ts
francoislg: I want to separate the `VisitorIdProvider` interface's property from the implementation.

If I rename `visitorId` to `currentVisitorId` and set it public, refactoring `currentVisitorId` also changes the interface's property.

PR #162, opened 2019-05-09
Applied round of code reviews following #160

francoislg:


src/client/analytics.ts
```diff
@@ -46,21 +48,28 @@ export interface AnalyticsClient {
...
+            version: Version
+        };
+    }
+
+    private options: ClientOptions;
 
-    constructor(opts: ClientOptions) {
-        if (typeof opts === 'undefined') {
+    constructor(opts: Partial<ClientOptions>) {
+        if (!opts) {
```
samisayegh: I think the best/most-actionable use of a `throw` here would be to check that a token was passed. Atm, I could pass an empty object to the constructor to get past the `throw`, but that does not resolve the underlying issue. What do you think of this:
```
constructor(opts) {
    this.options = {...}
    assertTokenIsDefined();
}

function assertTokenIsDefined() {
    if (!this.options.token) {
        throw new Error('You have to pass at least a token to this constructor')
    }
}
```

src/client/analytics.ts
```diff
@@ -150,16 +159,20 @@ export class CoveoAnalyticsClient implements AnalyticsClient {
...
-        var headers: any = {
+    protected getHeaders(): Record<string, string> {
+        var headers: Record<string, string> = {
             'Content-Type': `application/json`
         };
-        if (this.token) {
-            headers['Authorization'] = `Bearer ${this.token}`;
+
+        const { token } = this.options;
+        if (token) {
```
samisayegh: I don't think we need the condition here. Let's just pass whatever the token is, whether it's an empty string or something else. We have validation in the constructor.

Smaller point and more of an opinion, but I think it's easier to debug when you see the Authorization header has no token, rather than when there is no Authorization header.

src/client/analytics.ts
```diff
@@ -150,16 +159,20 @@ export class CoveoAnalyticsClient implements AnalyticsClient {
...
 
     protected getRestEndpoint(): string {
-        return `${this.endpoint}/rest/${this.version}`;
+        const { endpoint, version } = this.options;
+        return `${endpoint}/rest/${version}`;
     }
 
-    protected getHeaders(): any {
-        var headers: any = {
+    protected getHeaders(): Record<string, string> {
```
samisayegh: Could this function be `private`?

PR #163, opened 2019-05-28
Allow to queue init after queuing sends

francoislg:
Just a little safeguard, in case you have multiple async scripts.

We pre-process `init` events from the queue, then dequeue the rest.

### DIFF
--- a/src/coveoua/browser.ts
+++ b/src/coveoua/browser.ts
@@ -32,7 +32,15 @@ global.coveoanalytics = analytics;
 // On normal execution this library should be loaded after the snippet execution
 // so we will execute the actions in the `q` array
 if (coveoua.q) {
-    coveoua.q.forEach((args: [string, any[]]) => handleOneAnalyticsEvent.apply(void 0, args));
+    const isInitEvent = (args: [string, any[]]) => args[0] === 'init';
+    const isNotInitEvent = (args: [string, any[]]) => isInitEvent(args);
+    const processEvent = (args: [string, any[]]) => handleOneAnalyticsEvent.apply(void 0, args);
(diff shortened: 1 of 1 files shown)


src/coveoua/browser.ts
```diff
@@ -32,7 +32,15 @@ global.coveoanalytics = analytics;
...
+    const isInitEvent = (args: [string, any[]]) => args[0] === 'init';
+    const isNotInitEvent = (args: [string, any[]]) => !isInitEvent(args);
+    const processEvent = (args: [string, any[]]) => handleOneAnalyticsEvent.apply(void 0, args);
+    coveoua.q
+        .filter(isInitEvent)
+        .forEach(processEvent);
+    coveoua.q
+        .filter(isNotInitEvent)
+        .forEach(processEvent);
 }
```
samisayegh: I think from a code stand-point, all looks good. That said, would the logic be easier to follow if we sorted the array, placing the `init` event first, and then looped through calling `processEvent`? e.g.

```
const initEvent = coveoua.q.find(args => args[0] === 'init'); // assuming only one init event is allowed.
const otherEvents = coveoua.q.filter(args => args[0] !== 'init');
const reorderedEvents = [initEvent, ...otherEvents];
reorderedEvents.forEach(args => handleOneAnalyticsEvent.apply(void 0, args));
```

If there should only be one init event sent, then the `find` will get the first one it finds and ignore the rest. If you want to support multiple `init`, then a filter would be better as in your approach.

francoislg: I took a hybrid approach, I liked the idea of a single forEach on the spread array :)

PR #168, opened 2019-12-17
Add fix for store that gets encoded multiple time

olamothe:
If you have different version of coveoua included in a single page, it seems that older version will encode values differently than newer version, so we can end up with a string when the `getItem` functionality was called on the store.

And then all kind of random stuff would start to break.

This tries to add some logic to Parse until we get back a valid object if possible, with a bail if everything is fubar.

### DIFF
--- a/src/history.spec.ts
+++ b/src/history.spec.ts
@@ -47,11 +47,7 @@ test('HistoryStore should be able to retrieve most recent element in the history
     const storageInMemory = new TestMemoryStorage()
     historyStore = new history.HistoryStore(storageInMemory)
 
-    const doubleEncoded = JSON.stringify(JSON.stringify([{
-        name: 'name',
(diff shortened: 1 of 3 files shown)


olamothe:
I saw the problem arise while we are in the process of delivering MVP release for In Product. We include a newer version of coveoua on load of the In Product button to log a page view.

Then, when you open the popup and the JSUI is loaded, we get a "different" older version that is included directly in the JSUI source file, which is version 0.7.4.

And the local storage gets fubar when that happens.

olamothe:
After more investigation, the actual problem I saw was that for a particular client, they override the global JSON.stringify function with their own "custom" one which has a flawed implementation which double encodes stuff it should not.

I'll re-open a new PR with another approach to fix this problem.

PR #169, opened 2020-01-06
Update npm token; provide stand-alone library bundle

samisayegh:
I updated the npm token with the hopes that it will fix the 401 error.

One concern I have though is when I install the .tarball, the directory structure does not correspond to the main property in the package.json.

The dir structure is:

```
dist
-- dist
---- .d.ts files
-- coveoua.js
-- coveoua.js.map
```
whereas the main property is `dist/index.js`.

~I'm not sure if this is due to a quirk of building on windows or if it is a genuine problem.~

#### Edit

I think the change in the directory structure occurred after the move to webpack starting in this commit: https://github.com/coveo/coveo.analytics.js/commit/1074500a72b8e71117e42a8a74f806f3ad9a3e30

I made a couple of changes in the build configuration so that people installing the package through npm can import the modules without the browser CDN logic. Main things are:

* Generate a separate bundle `library.js` with the exported modules.
* Adjust the `main` property in package.json to this file.
* Rename the dir storing the typedefinitions, and reference them via the `types` property in package.json


package.json
```diff
@@ -2,8 +2,9 @@
   "name": "coveo.analytics",
   "version": "1.0.0",
```
francoislg: Should the version also be updated?

olamothe: Yes, most probably.

It guess it should be updated to w/e the latest tag is right now.

samisayegh: Yep, but I prefer to do this in a separate dedicated commit. I will open a second PR momentarily,

PR #170, opened 2020-01-13
Release v1.1.0

samisayegh:


samisayegh:
I unpublished the package because when I download and import a class from it, I get a empty object. I have not been able to find out why yet, but I suspect it has to do with the webpack config.

PR #172, opened 2020-02-03
Kind of support the `ec:addProduct` action

francoislg:
Changelog:

* Can use the `ec` plugin which uses hooks to change stuff
* Split payload enhancer into hooks
* Created a very simple "params" container

I wanted to get some feedback on the direction this is going. 

For now, each product is stored in a local array and will be appended to the payload when the `pageview` event is sent.

I borrowed many things from `commerce-ca-js` and adapted some others :)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -7,7 +7,7 @@ O=o.getElementsByTagName(v)[0];O.parentNode.insertBefore(u,O)
 
 // Replace YOUR-TOKEN with your real token
 // (eg: an API key which has the rights to write into Coveo UsageAnalytics)
-coveoua('init','YOUR-TOKEN');
-coveoua('ec:addProduct', {wow: "cool"});
+coveoua('init','YOUR-TOKEN', 'https://usageanalyticsdev.coveo.com');
+coveoua('ec:addProduct', {name: "wow", id: 'something', brand: 'brand', custom: 'ok'});
 coveoua('send', 'pageview');
 </script>
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -6,7 +6,6 @@ import {
     ClickEventResponse,
     CustomEventRequest,
     CustomEventResponse,
-    EventBaseRequest,
     EventType,
     HealthResponse,
     SearchEventRequest,
@@ -14,12 +13,14 @@ import {
     ViewEventRequest,
     ViewEventResponse,
     VisitResponse,
-    IRequestPayload
+    IRequestPayload,
 } from '../events';
 import { HistoryStore } from '../history';
 import { VisitorIdProvider } from './analyticsRequestClient';
-import { WebStorage, getAvailableStorage, CookieStorage } from '../storage';
+import { WebStorage, CookieStorage } from '../storage';
 import { hasLocalStorage, hasCookieStorage } from '../detector';
+import { addDefaultValues } from '../hook/addDefaultValues';
+import { enhanceViewEvent } from '../hook/enhanceViewEvent';
 
 export const Version = 'v15';
 
@@ -35,6 +36,8 @@ export interface ClientOptions {
     version: string;
 }
 
+export type AnalyticsClientSendEventHook = <T>(eventType: string, payload: any) => T;
+
 export interface AnalyticsClient {
     sendEvent(eventType: string, payload: any): Promise<AnyEventResponse | void>;
     sendSearchEvent(request: SearchEventRequest): Promise<SearchEventResponse | void>;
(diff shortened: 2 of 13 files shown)


src/client/analytics.ts
```diff
@@ -188,27 +207,6 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         store.addElement(historyElement);
     }
 
-    private getPayloadForTypeOfEvent(eventType: EventType, payload: any) {
```
francoislg: This code here was split into the two hooks

src/coveoua/simpleanalytics.ts
```diff
@@ -31,6 +33,10 @@ export class CoveoUA {
             });
         } else if (typeof token === 'object' && typeof token.sendEvent !== 'undefined') {
             this.client = token;
+        }
+
+        if (this.client) {
+            this.plugins.register('ec', new EC({client: this.client, params: this.params}));
```
francoislg: Always register the plugin for now, but we could implement `require` in the future to match the spec

src/coveoua/simpleanalytics.ts
```diff
@@ -90,23 +74,26 @@ export class CoveoUA {
...
-/** @deprecated */
-export const SimpleAPI = CoveoUA;
 
 export const coveoua = new CoveoUA();
 
-export const handleOneAnalyticsEvent = (action: string, ...params: any[]): any => {
-    const actionFunction = (<any>coveoua)[action];
-    if (actionFunction) {
+export const handleOneAnalyticsEvent = (command: string, ...params: any[]): any => {
+    const [, trackerName, pluginName, fn ] = /^(?:(\w+)\.)?(?:(\w+):)?(\w+)$/.exec(command)!;
```
francoislg: Regex and extraction were taken from `tracker.js` in the `static run()` method, works well so far 😄

public/index.html
```diff
@@ -7,7 +7,7 @@
 
 // Replace YOUR-TOKEN with your real token
 // (eg: an API key which has the rights to write into Coveo UsageAnalytics)
-coveoua('init','YOUR-TOKEN');
-coveoua('ec:addProduct', {wow: "cool"});
+coveoua('init','YOUR-TOKEN', 'https://usageanalyticsdev.coveo.com');
```
dlafreniere: should it be `create` to match with the ga syntax?
https://developers.google.com/analytics/devguides/collection/analyticsjs/command-queue-reference#create

src/hook/enhanceViewEvent.ts
```diff
@@ -0,0 +1,16 @@
+import { AnalyticsClientSendEventHook } from '../client/analytics';
+import { ViewEventRequest } from '../events';
+
+export const enhanceViewEvent: AnalyticsClientSendEventHook = (
+    eventType,
+    payload
+) => {
+    return eventType === 'view'
```
dlafreniere: use EventType constant?

src/plugins/ec.ts
```diff
@@ -0,0 +1,58 @@
+import { AnalyticsClient } from '../client/analytics';
+import { Params, Product } from '../coveoua/params';
+import { EventType } from '../events';
+
+const productKeysMapping: {[name: string]: string} = {
```
dlafreniere: for more information:
https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#enhanced-ecomm

src/coveoua/params.ts
```diff
@@ -0,0 +1,35 @@
+export interface AllParams {
+    products: Product[];
+    [name: string]: any;
+}
+
+export interface Product {
```
dlafreniere: eventually, theses could help:
https://github.com/laxersnine/ga-enhanced-ec-note#object

public/index.html
francoislg: `init` does not quite match the `create` syntax

`ga('create', [trackingId], [cookieDomain], [name], [fieldsObject]);`, but our implementation is:

`coveoua("init", [token | client], [endpointUrl])`.

so I would keep the `init` name until we get closer to the `create` one

src/coveoua/simpleanalytics.ts
toussaints: Wouldn't that be breaking?

francoislg: yeah, but this branch merges into `measurement-protocol`

bigluck: let's go on with the `ec` plugin inside the library ;)

PR #173, opened 2020-02-04
Support parsing variable arguments for specific event types

francoislg:
The signature is defined here: https://developers.google.com/analytics/devguides/collection/analyticsjs/events#implementation

I tried multiple things to not disturb too much the code, I feel like the simple configuration allows mapping the arguments in the right order quite easily

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -14,8 +14,8 @@ import {
     ViewEventResponse,
     VisitResponse,
     IRequestPayload,
+    VariableArgumentsPayload,
 } from '../events';
-import { HistoryStore } from '../history';
 import { VisitorIdProvider } from './analyticsRequestClient';
 import { WebStorage, CookieStorage } from '../storage';
 import { hasLocalStorage, hasCookieStorage } from '../detector';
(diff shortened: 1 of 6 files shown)


src/events.ts
```diff
@@ -14,6 +14,7 @@ export interface SearchDocument {
 }
 
 export type SendEventArguments = [EventType, ...any[]];
+export type VariableArgumentsPayload = [any] | [string, any] | [string, string, any] | [string, string, string, any] | [string, string, string, string, any];
```
dlafreniere: the ga typings do some interesting things for this. They seem to explicitly define the combinations they want to support
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/google.analytics/index.d.ts#L543

src/client/analytics.ts
```diff
@@ -193,18 +202,25 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
 
-    private addPageViewToHistory(pageViewValue: string) {
-        const store = new HistoryStore();
-        const historyElement = {
-            name: 'PageView',
-            value: pageViewValue,
-            time: JSON.stringify(new Date()),
-        };
-        store.addElement(historyElement);
+    private parseVariableArgumentsPayload(fieldsOrder: string[], payload: VariableArgumentsPayload) {
```
toussaints: dat method 😂

francoislg: I really tried

PR #174, opened 2020-02-05
Add ec:setAction and encapsulate EC data in the plugin

francoislg:
So I had a massive misunderstanding in the previous code :rip:

Exploring the `ca.js` code, I thought `this.data` was the same in `tracker.js` as in `ec.js`, but they are *different instances* of the params object 🤦‍♂ 

This means we can 100% encapsulate EC data within the plugin, which makes the whole thing way more simple

Also, I fixed two things reported by Luca (🙇 thanks again):

* Clearing the EC data when an event is sent with them
* The product index starts at 1, not 0

Next step is to start implementing default context information that are appended to all events 🎊


src/plugins/ec.ts
```diff
@@ -15,36 +14,71 @@ const productKeysMapping: {[name: string]: string} = {
...
-        this.params = params;
 
         this.addHooksForPageView();
         this.addHooksForEvent();
+        this.addHooksForECEvents();
     }
 
     addProduct(product: Product) {
-        this.params.products = [...this.params.products, product];
+        this.products = [...this.products, product];
```
bigluck: `this.products.push(product)` ?

francoislg:
I forgot to not merge `mp-defaultcontext`, so you should be able to use this link to get the commits from the original content of this PR: https://github.com/coveo/coveo.analytics.js/pull/174/files/e7235da63547c9ba07613ddb7244e2eace880012

src/hook/addDefaultValues.ts
```diff
@@ -1,9 +1,12 @@
 import { AnalyticsClientSendEventHook } from '../client/analytics';
+import { EventType } from '../events';
+
+const eventTypesForDefaultValues: string[] = [EventType.click, EventType.custom, EventType.search, EventType.view];
 
 export const addDefaultValues: AnalyticsClientSendEventHook = (eventType, payload) => {
-    return {
+    return eventTypesForDefaultValues.indexOf(eventType) !== -1 ? {
```
dlafreniere: `includes`?

francoislg: https://caniuse.com/#search=includes

Unfortunately, no `includes` in IE11 😢 The tradeoff is not worth it, IMO.

src/plugins/ec.ts
```diff
@@ -15,57 +14,146 @@ const productKeysMapping: {[name: string]: string} = {
...
+            variableLengthArgumentsNames: ['eventCategory', 'eventAction', 'eventLabel', 'eventValue'],
+            addDefaultContextInformation: true,
         });
     }
 
-    private enhancePayload(payload: any) {
-        return this.params.products.reduce((newPayload, product, index) => {
+    private addECDataToPayload(payload: any) {
+        const payloadWithConvertedKeys = this.convertKeysToMeasurementProtocol({
+            action: this.action,
```
dlafreniere: this was different

PR #177, opened 2020-02-06
Unit tests for analytics and EC

francoislg:
It already found some bugs:

* action was set even if the value was "undefined"
* VariableArgumentsPayload type didn't support "no additional argument" like `send("send", "x")`
* sendEvent didn't allow `string`, so you could not use the API with a custom event type, such as the ones supported in `ec`.

All of those didn't block JS-code from calling them properly, but now they all have better typings.

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -48,6 +48,21 @@ describe('ec events', () => {
         });
     });
 
+    it('can send a pageview event with options', async () => {
+        coveoua('init', aToken, anEndpoint);
+        await coveoua('send', 'pageview', 'page', {
+            title: 'wow',
+            location: 'http://right.here',
+        });
+
+        assertRequestSentContainsEqual({
+            ...defaultContextValues,
+            page: 'page',
+            title: 'wow',
+            location: 'http://right.here'
+        });
+    });
+
     const assertRequestSentContainsEqual = (toContain: {[name: string]: any}) => {
         expect(fetchMock.called()).toBe(true);
 
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -1,0 +1,62 @@
+import 'isomorphic-fetch';
+import * as fetchMock from 'fetch-mock';
+import { DefaultEventResponse } from '../src/events';
+import coveoua from '../src/coveoua/browser';
+
+describe('ec events', () => {
+    const aToken = 'token';
+    const anEndpoint = 'http://bloup';
+    const aVisitorId = '123';
+
+    const defaultContextValues = {
+        cid: '',
+        dl: `${location.protocol}//${location.hostname}${location.pathname.indexOf('/') === 0 ? location.pathname : `/${location.pathname}`}${location.search}`,
+        sr: `${screen.width}x${screen.height}`,
+        sd: `${screen.colorDepth}-bit`,
+        ul: navigator.language,
+        ua: navigator.userAgent,
+        dr: document.referrer,
+        dt: document.title,
+        de: document.characterSet,
+    };
+
+    beforeEach(() => {
+        const address = `${anEndpoint}/rest/v15/analytics/collect`;
+        const eventResponse: DefaultEventResponse = {
+            visitId : 'firsttimevisiting',
+            visitorId: aVisitorId
+        };
+        fetchMock.reset();
+        fetchMock.post(address, eventResponse);
+        fetchMock.post(`${address}?visitor=${aVisitorId}`, eventResponse);
+    });
+
+    it('can send a product detail view event', async () => {
+        coveoua('init', aToken, anEndpoint);
+        coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', custom: 'ok'});
+        coveoua('ec:setAction', 'detail', {storeid: 'amazing'});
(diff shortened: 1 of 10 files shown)


src/client/analytics.spec.ts
```diff
@@ -1,42 +1,191 @@
...
 
         const response = await client.sendViewEvent(viewEvent);
-        expect(response).toEqual(viewEventResponse);
+        expect(response).toEqual(eventResponse);
 
         expect(fetchMock.called()).toBe(true);
 
-        const [path, params] = fetchMock.lastCall();
-        expect(path).toBe(address);
+        const [path, { headers: headers, body }] = fetchMock.lastCall();
```
dlafreniere: `headers: headers`?

src/client/analytics.spec.ts
```diff
@@ -1,42 +1,191 @@
...
+            const parsedBody = JSON.parse(body.toString());
+            expect(parsedBody).toEqual({
+                [argumentNames[0]]: 'something',
+                [argumentNames[1]]: 'another',
+                [argumentNames[2]]: '🌈'
+            });
+        });
+
+        it('should properly handle a finished object argument', async () => {
+            await client.sendEvent(specialEventType, 'something', { mergethis: 'plz' });
```
dlafreniere: would it be more useful to use the example calls that you showed us from the ga docs? (hitType, eventValue, etc)?

src/client/analytics.ts
```diff
@@ -245,16 +245,19 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
             clientId: this.visitorId,
             location: `${location.protocol}//${location.hostname}${location.pathname.indexOf('/') === 0 ? location.pathname : `/${location.pathname}`}${location.search}`,
-            referrer: document.referrer,
             screenResolution: `${screen.width}x${screen.height}`,
             screenColor: `${screen.colorDepth}-bit`,
-            encoding: document.characterSet,
             language: navigator.language,
-            title: document.title,
             userAgent: navigator.userAgent,
+            ...documentContext
```
dlafreniere: `...locationContext`
`...navigatorContext`
?

src/plugins/ec.spec.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    it('should convert known product keys into the measurement protocol format', () => {
+        ec.addProduct({ 'name': '🧀', price: '5.99$' });
+
+        const result = executeRegisteredHook(ECPluginEventTypes.event, {});
+
+        expect(result).toEqual({ 'pr1nm' : '🧀', 'pr1pr': '5.99$' });
+    });
+
+    it('should allow adding multiple products', () => {
```
dlafreniere: 🥇

src/client/analytics.spec.ts
```diff
@@ -1,42 +1,191 @@
...
+        const [, { body }] = fetchMock.lastCall();
+
+        const parsedBody = JSON.parse(body.toString());
+        expect(parsedBody).toMatchObject({
+            language: 'en',
+            userAgent: navigator.userAgent
+        });
+    });
+
+    it('should send the events in the order they were called', async () => {
```
francoislg: I dedicate this test to @bigluck

src/client/analytics.spec.ts
francoislg: my bad, will remove

functional/ec-events.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+        await coveoua('send', 'pageview');
+
+        assertRequestSentContainsEqual({
+            ...defaultContextValues,
+            pr1nm: 'wow',
+            pr1id: 'something',
+            pr1br: 'brand',
+            pr1custom: 'ok',
+            pa: 'detail',
+            storeid: 'amazing',
```
bigluck: I don't like the idea that users can break the protocol by pushing their custom parameters (or by overwriting the default params).
This is something we have absolutely to prevent; for this purpose, users can use Custom Dimensions, as described here: https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets

src/client/analytics.spec.ts
bigluck: 🙂

functional/ec-events.spec.ts
francoislg: My current stance is that if we need to filter out custom attributes, it means that we must create a whitelist of accepted parameters. I would rather have this whitelist in the backend, since it would be easier to update.

I would assume that our user will follow the spec as to what attributes should be sent, for now.

I am open to revisiting my stance, even more so if we have usage data to back it up 😄

bigluck: Nope, that's not a good solution.

First of first: the server has never to reject an event.
If we go in this direction, a client can break their integration by passing an unexpected param or a list of properties, and this is not acceptable.

Second point: the protocol has to be concerned with the Analytics team, the one responsable of collect, remap and then analyze all the parameters received from our clients.

Third point: we are going in the direction of normalizing the data, to standardize all the data injection, processing, and analysis. For this reason, is important, now, follow the standard rules.

For this first release we have to follow the measurement protocol; only after having added support of all the requirements, we can start to discuss adding custom plugins based on our advanced needs.

Trust me ;)

PR #178, opened 2020-02-10
Add time and unique eventId

francoislg:
Pretty simple change, added two context values.

We are getting pretty close to the full addToCart process :)

### DIFF
--- a/src/client/analytics.spec.ts
+++ b/src/client/analytics.spec.ts
@@ -43,7 +43,7 @@ describe('Analytics', () => {
 
         expect(fetchMock.called()).toBe(true);
 
-        const [path, { headers: headers, body }] = fetchMock.lastCall();
+        const [path, { headers, body }] = fetchMock.lastCall();
         expect(path).toBe(endpointForEventType(EventType.view));
 
         const h = headers as Record<string, string>;
@@ -105,7 +105,7 @@ describe('Analytics', () => {
 
     describe('with event type mapping with variable arguments', () => {
         const specialEventType = '🌟special🌟';
-        const argumentNames = ['1️⃣', '2️⃣', '3️⃣'];
+        const argumentNames = ['eventCategory', 'eventAction', 'eventLabel', 'eventValue'];
         beforeEach(() => {
             mockFetchRequestForEventType(EventType.custom);
             client.addEventTypeMapping(specialEventType, {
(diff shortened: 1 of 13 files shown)


src/client/analytics.ts
```diff
@@ -261,14 +262,20 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             language: navigator.language,
             userAgent: navigator.userAgent,
         };
-        return {
+        const eventContext = {
             clientId: this.visitorId,
```
bigluck: the visitor_id value need to become an optimistic value, instead of fetching it by calling the /visit endpoint, it has to be set by the client by storing a uuidv4 in the client's browser (in the next commits 🙂)

src/client/analytics.ts
```diff
@@ -261,14 +262,20 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+            time: new Date().valueOf().toString(),
+            eventId: uuidv4(),
```
dlafreniere: I'm confused? wouldn't the eventId be generated in the response?

src/plugins/ec.ts
```diff
@@ -45,7 +45,9 @@ const contextInformationMapping: {[key in keyof DefaultContextInformation]: stri
     screenResolution: 'sr',
     title: 'dt',
     userAgent: 'ua',
-    language: 'ul'
+    language: 'ul',
+    eventId: 'z',
```
dlafreniere: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#z

src/client/analytics.spec.ts
```diff
@@ -178,14 +174,31 @@ describe('Analytics', () => {
...
+            expect(second.eventId).not.toBe(third.eventId);
+            expect(third.eventId).not.toBe(first.eventId);
         });
 
         const assertHaveAllProperties = (object: any, properties: string[]) => {
             properties.forEach(property => expect(object).toHaveProperty(property));
         };
     });
+
+    const getParsedBodyCalls = (): any[] => {
```
toussaints: nice helper

src/client/crypto.ts
```diff
@@ -0,0 +1,14 @@
+
+export const uuidv4 = (a?: number | string): string => {
```
toussaints: have we taken that somewhere?

francoislg: From `ca.js`, with a TypeScript twist

src/plugins/ec.ts
toussaints: Not sure how that works either

francoislg: In the CA migration doc, `z` is used to send an event unique ID, not sure about its actual usefulness

@bigluck ?

francoislg:
Same mistake as yesterday, I merged in the wrong order 🤦‍♂ 

See here for actual changes: https://github.com/coveo/coveo.analytics.js/pull/178/commits/5b82900c9c6d6e4cfed9877906964079d07ecb9a

src/plugins/ec.ts
bigluck: It can be very useful for future automated tests we could set up to check the end-to-end implementation (from the client to our DB).
Unfortunately, beacon requests can't be intercepted after the event has been emitted, so there is no easy way to know which event_id was assigned by the server for a given event.
By reverting the ownership of the event_id, we can always know (and log) the event_id of the event emitted by the client, and then compare the data sent with the data received by our infrastructure, and even further with the final data remapped by our ELTs.

PR #179, opened 2020-02-10
Add pageViewId to EC events according to spec

francoislg:
I followed the document that Luca made for us.

What I understood, is that we must change the PageViewID if a second pageview is sent.

If you are on a traditional site, each page will generate a new UUID, this is fine.

But if you are on a Simple Page App, you should be able to send a "pageview" on each page change. Thus, it follows those rules:

* Any event sent will keep the first UUID generated. So three calls [event, pageview, event] should send the same UUID, since it is the same page.
* As soon as a second pageview occurs, we change the UUID like this: [event, pageview, event, pageview (change!), event, pageview (change!)].

Hope it makes sense :D

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -48,6 +48,39 @@ describe('ec events', () => {
         });
     });
 
+    it('can send a pageview event with options', async () => {
+        coveoua('init', aToken, anEndpoint);
+        await coveoua('send', 'pageview', 'page', {
+            title: 'wow',
+            location: 'http://right.here',
+        });
+
+        assertRequestSentContainsEqual({
+            ...defaultContextValues,
+            page: 'page',
+            title: 'wow',
+            location: 'http://right.here'
+        });
+    });
+
+    it('should change the pageViewId only when sending a second page view event', async () => {
+        coveoua('init', aToken, anEndpoint);
+        await coveoua('send', 'event');
+        await coveoua('send', 'event');
+        await coveoua('send', 'pageview');
+        await coveoua('send', 'event');
+        await coveoua('send', 'pageview');
+        await coveoua('send', 'event');
+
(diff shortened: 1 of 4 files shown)


src/plugins/ec.ts
```diff
@@ -162,4 +171,15 @@ export class EC {
             };
         }, {});
     }
+
+    private getPageViewId(eventType: string) {
```
bigluck: What about?
```
    constructor(...) {
        ...
        this.pageViewId = this.uuidGenerator();
```
and ...
```
    private getPageViewId(eventType: string) {
        if (eventType === ECPluginEventTypes.pageview) {
            if (this.hasSentFirstPageView) {
                this.pageViewId = this.uuidGenerator();
            }
            this.hasSentFirstPageView = true;
        }
        return this.pageViewId;
    }
```

src/plugins/ec.ts
```diff
@@ -19,6 +20,7 @@ const eventKeysMapping: {[name: string]: string} = {
     eventAction: 'ea',
     eventLabel: 'el',
     eventValue: 'ev',
+    pageViewId: 'a',
```
dlafreniere: I don't see this in the measurement protocol... 🤔

functional/ec-events.spec.ts
```diff
@@ -59,4 +92,8 @@ describe('ec events', () => {
             [key]: toContain[key]
```
francoislg: Holy shit, very subtle typo on line 58, `Object.keys((key: string) =>` should read `Object.keys(toContain).map((key: string) =>`

src/plugins/ec.ts
francoislg: According to Luca, this is the `adsense_id` parameter that is reused to get a page view ID

PR #180, opened 2020-02-11
Create visitorId before the first request is sent

francoislg:
Instead of relying on the first call to get a visitorId, we can automatically generate one when the script is loaded.

It will be set in the Cookies as usual, so the JSUI should read this one


bigluck:
I'm still investigating how this behavior can break our internal workflows (ETL/ELT). I need to chat with the Analytics team.

PR #181, opened 2020-02-11
Added missing hitType (t) parameter

francoislg:
So, I tried the collect endpoint in dev with all the recent changes and it was missing the `t` parameter.

There it is! I also found out that if you don't parse the response after `fetch`, the response stays empty in Chrome! Weird feature!

### DIFF
--- a/src/client/analyticsFetchClient.ts
+++ b/src/client/analyticsFetchClient.ts
@@ -39,6 +39,7 @@ export class AnalyticsFetchClient implements AnalyticsRequestClient {
 
             return visit;
         } else {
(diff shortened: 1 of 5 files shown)


src/client/analyticsFetchClient.ts
```diff
@@ -39,6 +39,7 @@ export class AnalyticsFetchClient implements AnalyticsRequestClient {
 
             return visit;
         } else {
+            try { response.json(); } catch { /* If you don't parse the response, it won't appear in the network tab. */ }
```
toussaints: 😦

bigluck: hahaha

PR #182, opened 2020-02-11
Implement dl and dr logic, bring context into EC plugin

francoislg:
This one has more substantial changes:

* The third argument is used to explicitly pass a page name: `coveoua("send", "pageview", "/PAGE")`, so I brought back `ca.js` logic to properly update the location using the page.
* Subsequent pageview event should use the previous pageview location as their referrer. This is useful for SPA.
* I brought back the `contextInformation`, because there was more complicated logic to it, it required direct access to the `ec` state to properly infer the values, depending on `pageview` and everything. It was just less trouble to bring the whole thing.

### DIFF
--- a/src/client/location.ts
+++ b/src/client/location.ts
@@ -1,0 +1,1 @@
+export const getFormattedLocation = (location: Location) => `${location.protocol}//${location.hostname}${location.pathname.indexOf('/') === 0 ? location.pathname : `/${location.pathname}`}${location.search}`;
--- a/src/plugins/ec.spec.ts
+++ b/src/plugins/ec.spec.ts
@@ -5,12 +5,27 @@ describe('EC plugin', () => {
     let ec: EC;
     let client: ReturnType<typeof createAnalyticsClientMock>;
 
+    const someUUIDGenerator = jest.fn(() => someUUID);
     const someUUID = '13ccebdb-0138-45e8-bf70-884817ead190';
-    const defaultResult = { 'a': someUUID, t: ECPluginEventTypes.event };
+    const defaultResult = {
+        a: someUUID,
+        de: document.characterSet,
+        dl: 'http://localhost/',
+        dr: 'http://somewhere.over/therainbow',
+        dt: 'MAH PAGE',
+        sd: '24-bit',
+        sr: '0x0',
+        tm: expect.any(String),
+        ua: navigator.userAgent,
+        ul: 'en-US',
+        t: ECPluginEventTypes.event,
+        z: someUUID,
+    };
 
     beforeEach(() => {
+        jest.clearAllMocks();
         client = createAnalyticsClientMock();
-        ec = new EC({ client, uuidGenerator: () => someUUID });
+        ec = new EC({ client, uuidGenerator: someUUIDGenerator });
     });
 
     it('should register a hook in the client', () => {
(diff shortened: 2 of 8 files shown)


src/client/location.ts
```diff
@@ -0,0 +1 @@
+export const getFormattedLocation = (location: Location) => `${location.protocol}//${location.hostname}${location.pathname.indexOf('/') === 0 ? location.pathname : `/${location.pathname}`}${location.search}`;
```
dlafreniere: how does this differ from [`Location.toString`](https://developer.mozilla.org/en-US/docs/Web/API/Location)?

src/plugins/ec.ts
```diff
@@ -175,13 +181,61 @@ export class EC {
...
+    }
+
+    private resolveCurrentLocation(eventType: string, payload: any) {
+        const currentLocation = getFormattedLocation(window.location);
+        if (!payload.page || eventType !== ECPluginEventTypes.pageview) {
+            return currentLocation;
+        }
+
+        const newPage = payload.page.replace(/^\/?(.*)$/, '/$1');
+        return `${currentLocation.split(/\//).slice(0, 3).join('/')}${newPage}`;
```
dlafreniere: :wut:

src/plugins/ec.ts
```diff
@@ -175,13 +181,61 @@ export class EC {
...
+        const screenContext = {
+            screenResolution: `${screen.width}x${screen.height}`,
+            screenColor: `${screen.colorDepth}-bit`,
+        };
+        const navigatorContext = {
+            language: navigator.language,
+            userAgent: navigator.userAgent,
+        };
+        const eventContext = {
+            time: new Date().valueOf().toString(),
```
dlafreniere: `Date.now()`?

src/client/location.ts
francoislg: `location.toString()` is somehow not brought up by TypeScript, so I did not use it for this reason 🤔

src/plugins/ec.ts
francoislg: I pulled this from `ca.js`.

So 

`"https://github.com/coveo/coveo.analytics.js/pull/182".split("/")` gives:

`["https:", "", "github.com", "coveo", "coveo.analytics.js", "pull", "182"]` so `slice(0, 3).join("/")` gives `https://github.com`.

Let me come up with something clearer 😄

src/client/location.ts
dlafreniere: I think `.href` returns the same thing as `toString`

src/plugins/ec.spec.ts
```diff
@@ -144,6 +159,30 @@ describe('EC plugin', () => {
         });
     });
 
+    it('should call the uuidv4 method', async () => {
+       await executeRegisteredHook(ECPluginEventTypes.event, {});
+       await executeRegisteredHook(ECPluginEventTypes.event, {});
+       await executeRegisteredHook(ECPluginEventTypes.event, {});
+
+       // One for generating pageViewId, one for each individual event.
+        expect(someUUIDGenerator).toHaveBeenCalledTimes(1 + 3);
```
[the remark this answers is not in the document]
francoislg: more expressive than a simple `4`, but I agree the comment is enough

PR #183, opened 2020-02-12
Added generic parameter override

francoislg:
So I wanted to allow setting `userId` and potentially other keys that could be overridden by our clients.

However, the `params` had to be injected *after* the context keys were added, but also should be converted to the measurement protocol.

This is why I migrated the measurement protocol translating into the client.

Products were a little trickier to "translate", so I left them in the `ec` plugin :/ 

I also improved the `ec-events` spec to be a little stricter on the content returned using matchers, and it now also simulates getting back the `visitorId` 💪

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -4,9 +4,12 @@ import { DefaultEventResponse } from '../src/events';
 import coveoua from '../src/coveoua/browser';
 
 describe('ec events', () => {
+    const initialLocation = `${window.location}`;
     const aToken = 'token';
     const anEndpoint = 'http://bloup';
-    const aVisitorId = '123';
+
+    const numberFormat = /[0-9]+/;
+    const guidFormat = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
 
     const defaultContextValues = {
         dl: `${location.protocol}//${location.hostname}${location.pathname.indexOf('/') === 0 ? location.pathname : `/${location.pathname}`}${location.search}`,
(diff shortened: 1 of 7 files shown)


functional/ec-events.spec.ts
```diff
@@ -4,9 +4,12 @@ import { DefaultEventResponse } from '../src/events';
 import coveoua from '../src/coveoua/browser';
 
 describe('ec events', () => {
+    const initialLocation = `${window.location}`;
     const aToken = 'token';
     const anEndpoint = 'http://bloup';
-    const aVisitorId = '123';
+
+    const numberFormat = /[0-9]+/;
+    const guidFormat = /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/;
```
dlafreniere: at this point should we pull in a lib like uuid4?
https://github.com/thenativeweb/uuidv4

francoislg: I try to limit the shipped dependencies we include in this lib, simply because it brings the size up.

This is not complicated enough to justify a module IMHO.

src/client/measurementProtocolMapper.ts
```diff
@@ -0,0 +1,85 @@
...
+    coupon: 'tcc',
+    affiliation: 'ta',
+    step: 'cos',
+    option: 'col'
+};
+
+type DefaultContextInformation = ReturnType<typeof EC.prototype.getDefaultContextInformation> & ReturnType<typeof EC.prototype.getLocationInformation>;
+const contextInformationMapping: {[key in keyof DefaultContextInformation]: string} = {
+    hitType: 't',
+    pageViewId: 'pid',
```
francoislg: @dlafreniere your suggestion was accepted by the UA council

functional/ec-events.spec.ts
dlafreniere: I agree, I meant between this, and the code that generates UUIDs

src/client/measurementProtocolMapper.ts
```diff
@@ -0,0 +1,85 @@
+import { EC, Product } from '../plugins/ec';
```
toussaints: i like the idea of this file

PR #184, opened 2020-02-12
Spec for the exact addToCart event

francoislg:
A final one that acts as a very strong validation of the `addToCart` event according to the GA spec VS what is currently returned by the `ca.js` implementation.

I wanted to have the event as close as what the `ca` script was returning, so I kept everything and commented on the changes.

Let me know what you think! 

I consider this my final stamp of proof, so when this gets merged, I think we can consider the goal from this sprint achieved :)

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -161,9 +161,10 @@ describe('ec events', () => {
             'category': 'category',
             'brand': 'brand',
             'variant': 'variant',
-            'price': 'price',
-            'quantity': '1'
+            'price': 0,
+            'quantity': 0
         };
+        await coveoua('set', 'currencyCode', 'EUR');
         await coveoua('ec:addProduct', product);
         await coveoua('ec:setAction', 'add');
         await coveoua('send', 'event', 'UX', 'click', 'add to cart');
@@ -174,6 +175,7 @@ describe('ec events', () => {
         expect(event).toEqual({
             pid: expect.stringMatching(guidFormat), // Key changed from `a`
             cid: expect.stringMatching(guidFormat),
+            cu: 'EUR',
             de: defaultContextValues.de,
             dl: defaultContextValues.dl,
             dr: defaultContextValues.dr,
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -38,6 +38,7 @@ describe('ec events', () => {
                 visitorId,
             } as DefaultEventResponse;
         });
+        coveoua('reset');
         coveoua('init', aToken, anEndpoint);
     });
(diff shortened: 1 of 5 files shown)


functional/ec-events.spec.ts
```diff
@@ -152,6 +153,54 @@ describe('ec events', () => {
...
+    it('should be able to follow the complete addToCart flow', async () => {
+        // https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#add-remove-cart
+        const product = {
+            'id': 'id',
+            'name': 'name',
+            'category': 'category',
+            'brand': 'brand',
+            'variant': 'variant',
+            'price': 'price',
+            'quantity': '1'
```
bigluck: @francoislg  we have a bug here, the quantity, price and position params can have values like 0, which is a "false-like" value and the library remove it from the payload.
We need to be sure to pass these values, even if they are 0

functional/ec-events.spec.ts
```diff
@@ -152,6 +153,56 @@ describe('ec events', () => {
...
+            'quantity': 0
+        };
+        await coveoua('set', 'currencyCode', 'EUR');
+        await coveoua('ec:addProduct', product);
+        await coveoua('ec:setAction', 'add');
+        await coveoua('send', 'event', 'UX', 'click', 'add to cart');
+
+        const [event] = getParsedBody();
+
+        // Event directly extracted from `ca.js` with the same sequence of event.
```
dlafreniere: not sure if these comments will be useful for very long

src/client/analytics.ts
```diff
@@ -251,8 +251,9 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
     }
 
     private removeEmptyPayloadValues(payload: IRequestPayload): IRequestPayload {
+        const isNotEmptyValue = (value: any) => (typeof (value) !== 'undefined' && value !== null && value !== '');
```
dlafreniere: what about `[]` or `{}`, is it possible to pass them as values?

bigluck: only text/int/float values are accepted, everything else has to be cast to string

functional/ec-events.spec.ts
francoislg: I kind of liked them since it directly explains the differences with `ca.js` 🤔

would you rather remove them?

src/client/analytics.ts
francoislg: Since the clean up is done before the `measurementProtocol` step, I would keep `[]` and `{}` in the remove empty values, and maybe strip them out over there

PR #185, opened 2020-02-18
Add typescript types for Product

dlafreniere:
Taken from:
https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#product-data

### DIFF
--- a/src/client/measurementProtocolMapper.ts
+++ b/src/client/measurementProtocolMapper.ts
@@ -1,16 +1,16 @@
 import { EC, Product } from '../plugins/ec';
 
 // Based off: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#enhanced-ecomm
-const productKeysMapping: {[name: string]: string} = {
+const productKeysMapping: {[key in keyof Product]: string} = {
     id: 'id',
     name: 'nm',
     brand: 'br',
     category: 'ca',
     variant: 'va',
-    position: 'ps',
     price: 'pr',
     quantity: 'qt',
-    coupon: 'cc'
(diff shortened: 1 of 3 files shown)


src/plugins/ec.ts
```diff
@@ -11,10 +11,27 @@ export const ECPluginEventTypes = {
...
-export interface Product {
-    [name: string]: string;
+// From https://stackoverflow.com/a/49725198/497731
+type RequireAtLeastOne<T, Keys extends keyof T = keyof T> =
+    Pick<T, Exclude<keyof T, Keys>>
+    & {
+        [K in Keys]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<Keys, K>>>
+    }[Keys];
+
+export interface ProductProperties {
```
francoislg: Maybe don't export this one? 🤔 Just so that we don't get confused with `ProductProperties` VS `Product` types

src/client/measurementProtocolMapper.ts
```diff
@@ -1,16 +1,16 @@
 import { EC, Product } from '../plugins/ec';
 
 // Based off: https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#enhanced-ecomm
-const productKeysMapping: {[name: string]: string} = {
+const productKeysMapping: {[key in keyof Product]: string} = {
```
francoislg: 👍 really cool

src/plugins/ec.ts
dlafreniere: at first I didn't but it seems to be used implicitly:
image

PR #186, opened 2020-02-20
Add Search Impressions to EC

dlafreniere:
still need to add the integration test, but would like some reviews

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -203,6 +203,63 @@ describe('ec events', () => {
           });
     });
 
+    it('should be able to follow the complete addImpression flow', async () => {
+        // https://developers.google.com/analytics/devguides/collection/analyticsjs/enhanced-ecommerce#product-impression
+        const productImpression1 = {
+            'id': 'P12345',
+            'name': 'Android Warhol T-Shirt',
+            'category': 'Apparel/T-Shirts',
+            'brand': 'Google',
+            'variant': 'black',
+            'list': 'Search Results',
+            'position': 1
+        };
+        const productImpression2 = {
+            'id': 'P67890',
+            'name': 'YouTube Organic T-Shirt',
+            'category': 'Apparel/T-Shirts',
+            'brand': 'YouTube',
+            'variant': 'gray',
+            'list': 'Search Results',
+            'position': 2
+        };
+
+        await coveoua('ec:addImpression', productImpression1);
+        await coveoua('ec:addImpression', productImpression2);
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            pid: expect.stringMatching(guidFormat),
+            cid: expect.stringMatching(guidFormat),
+            de: defaultContextValues.de,
+            dl: defaultContextValues.dl,
+            dr: defaultContextValues.dr,
+            dt: defaultContextValues.dt,
+            il1nm: productImpression1.list,
+            il1pi1id: productImpression1.id,
+            il1pi1nm: productImpression1.name,
+            il1pi1ca: productImpression1.category,
+            il1pi1br: productImpression1.brand,
+            il1pi1va: productImpression1.variant,
+            il1pi1ps: productImpression1.position,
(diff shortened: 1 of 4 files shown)


src/plugins/ec.ts
```diff
@@ -32,10 +32,29 @@ export interface ProductProperties {
...
+    category?: string;
+    variant?: string;
+    position?: number;
+    price?: number;
+}
+
+export type Impression = RequireAtLeastOne<ImpressionProperties, 'id' | 'name'>;
+export type BaseImpression = Omit<Impression, 'list'>;
+export interface ImpressionList {
+    listName?: string;
```
francoislg: 🤔 c'est tu trop confusing juste `name` ici ?

src/plugins/ec.ts
```diff
@@ -111,9 +135,29 @@ export class EC {
...
+            const {list: listName, ...baseImpression} = impression;
+            const index = lists.findIndex(list => list.listName === listName);
+            if (index === -1) {
+                lists.push({listName: listName, impressions: [baseImpression]});
+            } else {
+                lists[index].impressions.push(baseImpression);
+            }
+            return lists;
+
+        }, new Array<ImpressionList>());
```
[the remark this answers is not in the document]
dlafreniere: ah it was the `as Impression[]` that I was missing!

src/plugins/ec.ts
```diff
@@ -111,9 +135,29 @@ export class EC {
             };
         }, {});
 
+        const impressionsByList = this.impressions.reduce((lists, impression) => {
+            const {list: listName, ...baseImpression} = impression;
+            const index = lists.findIndex(list => list.listName === listName);
```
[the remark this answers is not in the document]
dlafreniere: at first I thought I would need the index, but that is handled later with the reduce

src/plugins/ec.ts
dlafreniere: je pensais que oui parce que Impression a la base a un `name`, mais je ne suis pas contre

src/plugins/ec.spec.ts
```diff
@@ -36,78 +36,179 @@ describe('EC plugin', () => {
...
-            ...defaultResult,
-            'pr1nm' : '🍟',
-            'pr1pr': 1.99,
-            'pr2nm' : '🍿',
-            'pr2pr': 3,
-            'pr3nm' : '🥤' ,
-            'pr3pr': 2,
+            const result = executeRegisteredHook(ECPluginEventTypes.event, {});
+
+            expect(result).not.toEqual({});
```
toussaints: can't we expect what it'll be instead of what it shouldn't be?

src/plugins/ec.ts
```diff
@@ -32,10 +32,29 @@ export interface ProductProperties {
 
 export type Product = RequireAtLeastOne<ProductProperties, 'id' | 'name'>;
 
+export interface ImpressionProperties {
+    id?: string;
```
toussaints: noob question mais pourquoi tout est optional sur ça?

src/plugins/ec.spec.ts
dlafreniere: I just indented this, but there is already a test that is asserting the format is okay. The goal of this test is the 2nd assertion

functional/ec-events.spec.ts
```diff
@@ -203,6 +203,63 @@ describe('ec events', () => {
...
+            'position': 2
+        };
+
+        await coveoua('ec:addImpression', productImpression1);
+        await coveoua('ec:addImpression', productImpression2);
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
```
dlafreniere: Validated against ca.js, the only difference was a non-standard field `il1pi1listindex: 0,` which doesn't add anything (its baked into the props)

src/plugins/ec.ts
dlafreniere: id et nom sont là en option (pour que tu puisses spécifier les deux) mais, le `RequireAtLeastOne` va forcer que t'ailles au moins un des deux (tel que le spec indique)

PR #187, opened 2020-02-24
[COM-305] Add anonymizeIp option

toussaints:
See https://developers.google.com/analytics/devguides/collection/analyticsjs/ip-anonymization

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -162,7 +162,44 @@ describe('ec events', () => {
         expect(event).toEqual({
             ...defaultContextValues,
             t: 'pageview',
-            aip: true
+            aip: 1
+        });
+    });
+
+    it('should be able to set the anonymizeIp with "true" string', async () => {
+        await coveoua('set', 'anonymizeIp', 'true');
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            aip: 1
+        });
+    });
+
+    it('should be able to set the anonymizeIp to false', async () => {
+        await coveoua('set', 'anonymizeIp', false);
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+        });
+    });
+
+    it('should be able to set the anonymizeIp to false for anything in between', async () => {
+        await coveoua('set', 'anonymizeIp', 'pôtato');
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
         });
     });
 
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -191,7 +191,19 @@ describe('ec events', () => {
         });
     });
 
-    it('should be able to set the anonymizeIp to false for anything in between', async () => {
+    it('should be able to set the anonymizeIp to false for undefined value', async () => {
+        await coveoua('set', 'anonymizeIp');
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+        });
+    });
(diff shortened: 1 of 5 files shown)


src/plugins/ec.ts
```diff
@@ -211,6 +211,7 @@ export class EC {
         const eventContext = {
             time: Date.now().toString(),
             eventId: this.uuidGenerator(),
+            anonymizeIp: false
```
francoislg: 🤔 I think we can omit this, and only send the value when it has been set by the user

toussaints: Done in last commit

functional/ec-events.spec.ts
```diff
@@ -153,6 +153,19 @@ describe('ec events', () => {
...
+    it('should be able to set the anonymizeIp', async () => {
+        await coveoua('set', 'anonymizeIp', true);
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            aip: true
```
bigluck: @toussaints  GA remap truly values to 1 and falsy values to 0 to reduce the request payload size, is this something we can replicate?

bigluck: The major problem I've found is that the value is always passed as is to the backend server:
`coveoua('set', 'anonymizeIp', 'test')`
will send the `aip=test` value instead of `aip=1`

toussaints: > @toussaints GA remap truly values to 1 and falsy values to 0 to reduce the request payload size, is this something we can replicate?

I guess, I'll look into it.

> The major problem I've found is that the value is always passed as is to the backend server:
> `coveoua('set', 'anonymizeIp', 'test')`
> will send the `aip=test` value instead of `aip=1`

Another option would be to map false to 0 and anything else to 1 (or the other way around). I don't see any exception mechanism for validation in place and I think this might be something we would want to implement coherently across values.

src/client/analytics.ts
```diff
@@ -260,6 +262,16 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             }), {});
     }
 
+    private validateParams(payload: IRequestPayload) : IRequestPayload {
+        const { anonymizeIp, ...rest } = payload;
+
+        if (anonymizeIp !== undefined && (anonymizeIp === true || anonymizeIp === 'true')) {
```
bigluck: yeah, I think we are in the right direction, but it still buggy because it's too exclusive.
What about something like this?

```
function isTruthy(value) {
  return !['0', 'false', 'undefined', 'null', '{}', '[]', ''].includes(`${value}`.toLowerCase());
}
if (isTruthy(anonymizeIp)) {
  rest['anonymizeIp'] = 1;
}
```

toussaints: My gut feeling is that we should default to false

bigluck: this is only my point of view, to be honest: it could be better to have only aip=1 when the value was set to a true value (or something ~true) and never have the aip value if it is false or is missing, to reduce the request payload ;)

toussaints: what your describing here is the current comportment of the code (if you look at the tests)
true or 'true' => aip:1
anything else => nothing added to the body

I do believe that's right, I'll let the other comment about this

@dlafreniere @francoislg @cnguyen-coveo

bigluck: Umh, no it's the inverse of your implementation :) . You have a list of ~true values and you're setting 1 when the input value matches your list; In my implementation, I have a list of ~false values and I'm setting 1 when the input value is not in the list :)

toussaints: What i'm saying is that the current code is 
- setting aip => 1 only if true or 'true'
- does not set aip in the payload otherwise (empty, wrong value, etc)

I do believe this is the way to go, but if you all agree that a random value should activate the feature then I'll change it.

Also, I'm considering adding 1 along with true and 'true' as per a @francoislg's suggestion.

francoislg: 💭 For me, the argument boils down to "what is the intention".

Since the value is `false` by default, it is more likely that calling `coveoua("set", "anonymiseIp", "anything")` is to set `true` 🤔

toussaints: Alright, went with the consensus and reverted the condition.

PR #190, opened 2020-03-11
[COM-303] remove unknown params from payload

cnguyen-coveo:


src/client/analytics.ts
```diff
@@ -270,15 +278,36 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
     }
 
+    private removeUnknownParameters(payload: IRequestPayload): IRequestPayload {
+        const productSubKeys = productKeysMappingValues.join("|")
+        const impressSubKeys = impressionKeysMappingValues.join("|")
+        const productKeyRegex = new RegExp(`^(pr[0-9]+)(${productSubKeys})$`)
+        const impressionKeyRegex = new RegExp(`^((il[0-9]+pi[0-9]+)(${impressSubKeys}))|(il[0-9]+nm)$`)
+        const newPayload = Object.keys(payload)
+            .filter((key) => { 
+                if (productKeyRegex.test(key) || impressionKeyRegex.test(key) || measurementProtocolKeysMappingValues.indexOf(key) != -1){
```
francoislg: I think it would be cleaner if you put this whole condition in `measurementProtocolMapper`, in a way that you would only import one method and use it like: `if (isMeasurementPrototocolKey(key)) { }`.

It would also encapsulate all `regex` and `subKeys` into the `measurementProtocolMapper` module, it makes sense for them to be there :D

src/client/analytics.ts
```diff
@@ -270,15 +278,36 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+                    return {key: payload[key]}
```
francoislg: Since this is a `.filter`, this expression is downcasted to `true`, so you could just `return true` instead

functional/ec-events.spec.ts
```diff
@@ -217,6 +215,26 @@ describe('ec events', () => {
             aip: 1,
         });
     });
+    it('should remove unknown values', async () => {
```
francoislg: Maybe `keys` would be more representative than `values` ? 🤔 

And maybe it could be split in two tests? `should remove unknown product keys` and `should remove unknown set key`

And a missing extra line before `it(` 🙈

functional/ec-events.spec.ts
```diff
@@ -217,6 +215,26 @@ describe('ec events', () => {
...
+        const [body] = getParsedBody();
+
+        expect(body).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            pr1nm: 'wow',
+            pr1id: 'something',
+            pr1br: 'brand',
+        });
+        expect(body).not.toContain({
```
francoislg: `toEqual` already checks if there are extra keys, so I think you can remove this

src/client/measurementProtocolMapper.ts
```diff
@@ -136,3 +136,7 @@ const convertImpressionToMeasurementProtocol = (
         };
     }, {});
 };
+
+export const measurementProtocolKeysMappingValues: String[] = Object.keys(measurementProtocolKeysMapping).map(key => Object(measurementProtocolKeysMapping)[key]);
```
francoislg: I guess you added `Object()` here because there was a typing issue with `Object.keys` that outputs `string[]` instead of `keyof {MAPPING}`?

Devin added a utility for that with `keysOf` right in this file for this purpose 😄

francoislg:
👏 :gg: on first "real code" PR 😄

functional/ec-events.spec.ts
```diff
@@ -217,6 +215,26 @@ describe('ec events', () => {
...
         });
     });
+    it('should remove unknown values', async () => {
+        await coveoua('set', 'unknownParam', 'unknown');
+        await coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', custom: 'ok'});
+        await coveoua('send', 'pageview');
+
+        const [body] = getParsedBody();
+
+        expect(body).toEqual({
```
dlafreniere: I don't think this assertion is necessary here

src/client/analytics.ts
```diff
@@ -270,15 +278,36 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             );
     }
 
+    private removeUnknownParameters(payload: IRequestPayload): IRequestPayload {
+        const productSubKeys = productKeysMappingValues.join("|")
```
dlafreniere: `productSubKeysMatchGroup`

src/client/analytics.ts
```diff
@@ -270,15 +278,36 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        const impressSubKeys = impressionKeysMappingValues.join("|")
+        const productKeyRegex = new RegExp(`^(pr[0-9]+)(${productSubKeys})$`)
+        const impressionKeyRegex = new RegExp(`^((il[0-9]+pi[0-9]+)(${impressSubKeys}))|(il[0-9]+nm)$`)
```
dlafreniere: because the regexes don't change, they can be initialized outside of the function, but I wonder if we can also separate the actual "filtering" logic with the the object manipulation:

```
const newPayload = Object.keys(payload)
            .filter((key) => { 
                if (keyIsKnown(key))
                    return {key: payload[key]}
                } else {
                 ....
```

src/client/measurementProtocolMapper.ts
dlafreniere: is the `map` really necessary?

cnguyen-coveo: yes since we want the values and not the keys

functional/ec-events.spec.ts
cnguyen-coveo: kept the `.not.toContain` instead of `.toEqual` since it seemed more representative of the purpose of the new method

src/client/measurementProtocolMapper.ts
dlafreniere: in that case, `Object.entries`?

cnguyen-coveo: I get `Property 'entries' does not exist on type 'ObjectConstructor'` same error when i do `Object.values`

functional/ec-events.spec.ts
francoislg: I agree it is more representative, but this test is less robust 😄 

`pr1custom` could "not be contained" for other reasons than the cleanup, by inadvertence. `toEqual` ensure that we get the original event, and nothing more

src/client/measurementProtocolMapper.ts
cnguyen-coveo: `Object.values` can be used as of es2017 but i dont think we are currently using that version. See here: https://stackoverflow.com/questions/42966362/how-to-use-object-values-with-typescript/42967397

src/client/analytics.ts
```diff
@@ -279,14 +277,10 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
     private removeUnknownParameters(payload: IRequestPayload): IRequestPayload {
-        const productSubKeys = productKeysMappingValues.join("|")
-        const impressSubKeys = impressionKeysMappingValues.join("|")
-        const productKeyRegex = new RegExp(`^(pr[0-9]+)(${productSubKeys})$`)
-        const impressionKeyRegex = new RegExp(`^((il[0-9]+pi[0-9]+)(${impressSubKeys}))|(il[0-9]+nm)$`)
         const newPayload = Object.keys(payload)
             .filter((key) => { 
-                if (productKeyRegex.test(key) || impressionKeyRegex.test(key) || measurementProtocolKeysMappingValues.indexOf(key) != -1){
-                    return {key: payload[key]}
+                if (isMeasurementPrototocolKey(key)){
```
francoislg: I think you are missing a `prettier` pass on this, there should be a space before the `{` 🤔 

You can run `npm run lint`, and also install the Prettier extension in VSCode and run the `Format Document` command

src/client/measurementProtocolMapper.ts
```diff
@@ -137,6 +137,18 @@ const convertImpressionToMeasurementProtocol = (
...
+const impressSubKeysMatchGroup = impressionKeysMappingValues.join("|")
+const productKeyRegex = new RegExp(`^(pr[0-9]+)(${productSubKeysMatchGroup})$`)
+const impressionKeyRegex = new RegExp(`^((il[0-9]+pi[0-9]+)(${impressSubKeysMatchGroup}))|(il[0-9]+nm)$`)
+
+export const isMeasurementPrototocolKey = (key: string): boolean => {
+    if (productKeyRegex.test(key) || impressionKeyRegex.test(key) || measurementProtocolKeysMappingValues.indexOf(key) != -1) {
+        return true;
+    }
+    return false;
+}
```
francoislg:
```suggestion
export const isMeasurementPrototocolKey = (key: string): boolean => {
    return productKeyRegex.test(key) || impressionKeyRegex.test(key) || measurementProtocolKeysMappingValues.indexOf(key) != -1);
}
```

small detail: unnecessary if 😄

src/client/measurementProtocolMapper.ts
francoislg: ☝️ don't forget this one 😄

src/client/analytics.ts
```diff
@@ -270,15 +276,32 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             );
     }
 
+    private removeUnknownParameters(payload: IRequestPayload): IRequestPayload {
+        const newPayload = Object.keys(payload)
+            .filter((key) => { 
+                if (isMeasurementPrototocolKey(key)){
```
francoislg: Oh shit, I just noticed, pro-*toto*-col

https://www.youtube.com/watch?v=FTQbiNvZqaY

https://www.youtube.com/watch?v=htgr3pvBr-I

PR #193, opened 2020-04-17
Allow custom values in products, impressions, and root level

francoislg:
[COM-356]

[COM-356]: https://coveord.atlassian.net/browse/COM-356

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -44,8 +44,23 @@ describe('ec events', () => {
         coveoua('init', aToken, anEndpoint);
     });
 
+    it('can set custom data at the root level and send a page view event', async () => {
+        await coveoua("set", "custom", {
+            "verycustom": "value"
+        })
+        await coveoua('send', 'pageview');
+
+        const [body] = getParsedBody();
+
+        expect(body).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            verycustom: 'value',
+        });
+    });
+
     it('can send a product detail view event', async () => {
-        coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', custom: 'ok'});
+        coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', unknown: 'ok'});
         coveoua('ec:setAction', 'detail', {storeid: 'amazing'});
         await coveoua('send', 'pageview');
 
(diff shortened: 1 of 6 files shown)


functional/ec-events.spec.ts
```diff
@@ -216,26 +270,32 @@ describe('ec events', () => {
...
     });
 
-    it('should remove unknown measurment protocol keys', async () => {
+    it('should remove unknown measurement protocol keys', async () => {
         await coveoua('set', 'unknownParam', 'unknown');
         await coveoua('send', 'pageview');
 
         const [body] = getParsedBody();
 
-        expect(body).not.toContain({
```
francoislg, line 225: `toContain` does *not* work with objects, only arrays!

cnguyen-coveo: my bad 😬

src/client/measurementProtocolMapper.ts
```diff
@@ -143,15 +143,50 @@ const measurementProtocolKeysMappingValues = keysOf(measurementProtocolKeysMappi
...
+        isProductKey,
+        isImpressionKey,
+        isKnownMeasurementProtocolKey,
+        isCustomKey
+    ].some(test => test(key));
 };
+
+export const convertCustomMeasurementProtocolKeys = (data: {[name: string]: string | {[name:string] :string}}) => {
+    return keysOf(data).reduce((all, current) => {
+        const match = customProductKeyRegex.exec(current) || customImpressionKeyRegex.exec(current);
```
cnguyen-coveo, line 171: Are we adding custom fields to impression? Since in my story it only specifies general and product

src/plugins/ec.ts
```diff
@@ -32,6 +32,7 @@ export interface ProductProperties {
     quantity?: number;
     coupon?: string;
     position?: number;
+    custom?: {[key: string]: string};
```
dlafreniere, line 35: should it be `[key: string]: any`?

francoislg: What about `string | number | boolean` ?

src/client/measurementProtocolMapper.ts
francoislg: The step to support them was minimal, so I added them 😄 

It makes more sense to support this in all of the types

src/plugins/ec.ts
dlafreniere: sure, provided we don't have to/want to support custom json

src/plugins/ec.ts
```diff
@@ -22,6 +22,10 @@ type RequireAtLeastOne<T, Keys extends keyof T = keyof T> = Pick<T, Exclude<keyo
         [K in Keys]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<Keys, K>>>;
     }[Keys];
 
+export type CustomValues = {
```
dlafreniere, line 25: 👌

cnguyen-coveo:
Should `index.html` also be updated since it has a `ec:addProduct` with a custom field?

PR #195, opened 2020-04-29
Create basic searchpages client

olamothe:
This is a partial PR, to ensure everybody thinks this is the right approach, before I continue in this direction.

For headless, we want to completely encapsulate/blackbox analytics through coveo.analytics.

In JSUI, one thing we had already created was a huge object containing all the analytics cause that could be used by all the different components you have inside a standard search page. All this logic and information was never ported back to coveo.analytics. 

See : https://github.com/coveo/search-ui/blob/master/src/ui/Analytics/AnalyticsActionListMeta.ts

This PR creates a `CoveoSearchPageClient` which exposes a bit of "higher" level methods that we could call from our headless library.

`CoveoSearchPagesClient` would be the source of truth for all "standard search pages" events.

Inside headless (so, from the consumer point of view), we could use it like this : 

```
import {CoveoSearchPageClient} from 'coveo.analytics';

export const analytics = new CoveoSearchPageClient({
  getBaseMetadata: () => ({headlessVersion: "0.1"}),
   getSearchUID: () => store.getState().result.searchUID
   getSearchEventRequestPayload: () =>({
       queryText: store.getState().query.expression,
       responseTime: store.getState().result.executionTime,
       ...... etc.
   }),
})



class HeadlessSearchbox() {
  constructor() {....}

  public submit() {
      // dispatch needed redux actions
      // .... etc
     analytics.logSearchboxSubmit();
  }

}

```

Once again, this is partial, not all tests are there, and not all events.

### DIFF
--- a/src/client/searchPageClient.spec.ts
+++ b/src/client/searchPageClient.spec.ts
@@ -1,0 +1,65 @@
+import * as fetchMock from 'fetch-mock';
+import { CoveoSearchPageClient, SearchPageEvents } from './searchPageClient'
+
+describe('SearchPageClient', () => {
+
+    let client: CoveoSearchPageClient
+
+    beforeEach(() => {
+        client = initClient()
+        fetchMock.mock(/.*/, {
+            visitId: 'visit-id',
+            visitorId: 'visitor-id',
+        });
+    })
+
+
+    afterEach(() => {
+        fetchMock.reset();
+    })
+
+    const initClient = () => {
+        return new CoveoSearchPageClient({}, {
+            getBaseMetadata: () => ({ 'foo': 'bar' }),
+            getSearchEventRequestPayload: () => ({
+                queryText: 'queryText',
+                responseTime: 123,
+            }),
+            getSearchUID: () => 'my-uid'
+        })
+    }
+
+    const expectMatchPayload = (actionCause: SearchPageEvents, meta = {}) => {
+        const [, { body }] = fetchMock.lastCall();
+        const customData = { 'foo': 'bar', ...meta }
+        expect(JSON.parse(body.toString())).toMatchObject({
+            queryText: 'queryText',
+            responseTime: 123,
+            actionCause,
+            customData,
+        });
+    }
+
+    describe('it should send proper payload', () => {
+
+        it('for #interfaceLoad', async () => {
+            await client.logInterfaceLoad()
+            expectMatchPayload(SearchPageEvents.interfaceLoad);
+        })
+
+        it('for #interfaceChange', async () => {
+            await client.logInterfaceChange({
+                interfaceChangeTo: 'bob'
+            })
+
+            expectMatchPayload(SearchPageEvents.interfaceChange, { interfaceChangeTo: 'bob' })
+        })
+
+        it('for #didyoumeanAutomatic', async () => {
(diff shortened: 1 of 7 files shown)


src/client/searchPageClient.spec.ts
```diff
@@ -0,0 +1,65 @@
...
+        const customData = { 'foo': 'bar', ...meta }
+        expect(JSON.parse(body.toString())).toMatchObject({
+            queryText: 'queryText',
+            responseTime: 123,
+            actionCause,
+            customData,
+        });
+    }
+
+    describe('it should send proper payload', () => {
```
francoislg, line 43: I tend to prefer names that read well with `describe/it` 🤔

Since there is no setup specific to these, I'd rather have direct `it should send proper #interfaceLoad payload` and such

src/client/searchPageClient.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    public logSearchboxClear() {
+        return this.logSearchEvent(SearchPageEvents.searchboxClear);
+    }
+
+    public logSearchboxAsYouType() {
+        return this.logSearchEvent(SearchPageEvents.searchboxAsYouType);
+    }
+
+    public logBreadcrumbFacet<T extends FacetMetadata | FacetRangeMetadata | CategoryFacetMetadata>(metadata: T) {
```
francoislg, line 140: 🤔 Are you sure you want the `extends` ? It allows adding any custom value to the metadata like so:

```
{
    facetId: 'string',
    facetField: 'string',
    facetTitle: 'string',
    facetValue: 'string',
    custom: 'string'
}
```

might want `logBreadcrumbFacet(metadata: FacetMetadata | FacetRangeMetadata | CategoryFacetMetadata)` instead?

src/client/searchPageClient.ts
```diff
@@ -0,0 +1,192 @@
...
+    public logOmniboxAnalytics(meta: OmniboxSuggestionsMetadata) {
+        return this.logSearchEvent(SearchPageEvents.omniboxAnalytics, meta);
+    }
+
+    public logOmniboxFromLink(meta: OmniboxSuggestionsMetadata) {
+        return this.logSearchEvent(SearchPageEvents.omniboxFromLink, meta);
+    }
+
+    private logSearchEvent(eventType: SearchPageEvents, metadata?: Record<string, any>) {
+        const meta = { ...this.provider.getBaseMetadata(), ...metadata }
```
francoislg, line 165: You could just call this one `customData`, same for the one in `logClickEvent`

src/client/searchPageClient.ts
```diff
@@ -0,0 +1,192 @@
...
+
+    constructor(private opts: Partial<ClientOptions>, private provider: SearchPageClientProvider) {
+        this.coveoAnalyticsClient = new CoveoAnalyticsClient(opts)
+    }
+
+    public logInterfaceLoad() {
+        return this.logSearchEvent(SearchPageEvents.interfaceLoad);
+    }
+
+    public logInterfaceChange(metadata: { interfaceChangeTo: string }) {
```
francoislg, line 112: Even if very simple, I'd create an interface for this metadata as well. Same for `logResultsSort`

PR #196, opened 2020-04-29
Use rollup for bundler

olamothe:
For headless, we want to be able to import es modules, which allows us to do effective tree shaking and compile time bundle size optimization.

Webpack unfortunately currently does not support generating es modules.

This should pretty much be the same output structure as before, (with slightly different bundle code generation for browser and commonjs library).


rollup.config.js
```diff
@@ -0,0 +1,56 @@
...
+            port: 9001,
+            open: true,
+        }) : null
+    ]
+}
+
+
+const libraryUmd = {
+    input: './src/coveoua/library.ts',
+    output: umdConfig(resolve(__dirname, './dist/library.js')),
```
samisayegh, line 35: The config between library and browser is almost identical. You could specify the `output` as an array of objects that hold the file and format keys. e.g.
```
output: [
    umdConfig('dist/library.js'),
    umdConfig('dist/coveoua.js')
]
```
Was resolving absolute paths necessary? I've used relative paths without issues.

The main difference is in watch mode, rollup will build two files instead of one. For a project this size though, I think it will only add a second or two.

PR #197, opened 2020-05-05
Fix empty value + custom events support

olamothe:
* For search events, the `queryText` being empty is a valid input, and should not be removed.
* Added supports for trigger(s) event in SearchPageClient
* Added support for logging custom events in general in SearchPageClient

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -1,5 +1,5 @@
-import { AnalyticsBeaconClient } from './analyticsBeaconClient';
-import { AnalyticsFetchClient } from './analyticsFetchClient';
+import {AnalyticsBeaconClient} from './analyticsBeaconClient';
+import {AnalyticsFetchClient} from './analyticsFetchClient';
 import {
     AnyEventResponse,
     ClickEventRequest,
@@ -16,13 +16,13 @@ import {
     IRequestPayload,
     VariableArgumentsPayload,
 } from '../events';
-import { VisitorIdProvider } from './analyticsRequestClient';
-import { WebStorage, CookieStorage } from '../storage';
-import { hasLocalStorage, hasCookieStorage } from '../detector';
-import { addDefaultValues } from '../hook/addDefaultValues';
-import { enhanceViewEvent } from '../hook/enhanceViewEvent';
-import { uuidv4 } from './crypto';
-import { convertKeysToMeasurementProtocol, isMeasurementProtocolKey, convertCustomMeasurementProtocolKeys } from './measurementProtocolMapper';
+import {VisitorIdProvider} from './analyticsRequestClient';
+import {WebStorage, CookieStorage} from '../storage';
+import {hasLocalStorage, hasCookieStorage} from '../detector';
+import {addDefaultValues} from '../hook/addDefaultValues';
+import {enhanceViewEvent} from '../hook/enhanceViewEvent';
+import {uuidv4} from './crypto';
+import {convertKeysToMeasurementProtocol, isMeasurementProtocolKey, convertCustomMeasurementProtocolKeys} from './measurementProtocolMapper';
 
 export const Version = 'v15';
 
@@ -79,7 +79,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
     private analyticsFetchClient: AnalyticsFetchClient;
     private bufferedRequests: BufferedRequest[];
     private beforeSendHooks: AnalyticsClientSendEventHook[];
-    private eventTypeMapping: { [name: string]: EventTypeConfig };
+    private eventTypeMapping: {[name: string]: EventTypeConfig};
     private options: ClientOptions;
 
     constructor(opts: Partial<ClientOptions>) {
@@ -92,7 +92,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
             ...opts,
         };
 
-        const { token } = this.options;
+        const {token} = this.options;
 
         this.cookieStorage = new CookieStorage();
         this.visitorId = '';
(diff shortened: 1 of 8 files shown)


src/client/analytics.ts
```diff
@@ -263,10 +263,21 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return parsedArguments;
     }
 
-    private removeEmptyPayloadValues(payload: IRequestPayload): IRequestPayload {
+    private getValidEmptyValuesForEvent(evtType: EventType | string): string[] {
```
olamothe, line 266: At first I thought there would be other parameters, but with hindsight I think there's only `queryText` that requires this special treatment. Not 100% sure.

So this whole method might be a bit overkill. 

Anyway, tell me if you'd prefer just a simpler `string[]` directly in `removeEmptyPayloadValues` or something.

francoislg:
I think you need to run `npm run lint` to fix some of the spacing issues

olamothe:
> I think you need to run `npm run lint` to fix some of the spacing issues

Hum this seems to be formatting some other random files I haven't touched ( `ev-events-spec.ts` , `events.ts` )

src/client/analytics.ts
```diff
@@ -263,10 +263,21 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        }
+        return []
+    }
+
+    private removeEmptyPayloadValues(payload: IRequestPayload, eventType: EventType | string): IRequestPayload {
+        const keysThatCanBeEmpty = this.getValidEmptyValuesForEvent(eventType)
+
         const isNotEmptyValue = (value: any) => typeof value !== 'undefined' && value !== null && value !== '';
         return Object.keys(payload)
             .filter((key) => isNotEmptyValue(payload[key]))
```
francoislg, line 279: I think you could create a `this.isKeyAllowedEmpty(eventType, key)` (or something like that) and call it here in the filter instead of adding the "concat" thing? 🤔 

I feel it is simpler this way:

`(key) => this.isKeyAllowedEmpty(eventType, key) || isNotEmptyValue(payload[key]))`

src/client/analytics.ts
```diff
@@ -263,10 +263,21 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return parsedArguments;
     }
 
-    private removeEmptyPayloadValues(payload: IRequestPayload): IRequestPayload {
+    private getValidEmptyValuesForEvent(evtType: EventType | string): string[] {
+        switch (evtType) {
```
francoislg, line 267: It would be cleaner to keep this thing into a simple object:

```
const keysThatCanBeEmpty = {
   [EventType.search]: ['queryText']
};
return keysThatCanBeEmpty[evtType] || [];
```

src/searchPage/searchPageEvents.ts
```diff
@@ -57,6 +57,29 @@ export enum SearchPageEvents {
      * Identifies the search event that gets logged when a suggested search query is selected from a standalone searchbox.
      */
     omniboxFromLink = 'omniboxFromLink',
+    /**
```
ThibodeauJF, line 60: Is a documentation review necessary for those?

olamothe: They are copied directly from JSUI (so, already reviewed)

src/client/analytics.ts
```diff
@@ -263,10 +263,19 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return parsedArguments;
     }
 
-    private removeEmptyPayloadValues(payload: IRequestPayload): IRequestPayload {
+    private isKeyAllowedEmpty(evtType: EventType | string, key: string) {
+        const keysThatCanBeEmpty: Partial<Record<EventType | string, string[]>> = {
```
samisayegh, line 267: Is the `Partial` type needed here?

src/client/analytics.ts
```diff
@@ -263,10 +263,19 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return parsedArguments;
     }
 
-    private removeEmptyPayloadValues(payload: IRequestPayload): IRequestPayload {
+    private isKeyAllowedEmpty(evtType: EventType | string, key: string) {
```
samisayegh, line 266: Since we are accepting `string`, are we getting any type-checking benefits of also specifying `EventType`?

If we can be strict, let's extend `EventType`.
If we need to be less-strict, let's use `string` and simplify here, line 267 and 275.

src/searchPage/searchPageEvents.ts
```diff
@@ -57,6 +57,29 @@ export enum SearchPageEvents {
...
+     * Identifies the custom event that gets logged when a user action triggers a new query set in the effective query pipeline on the search page.
+     */
+    triggerQuery = 'query',
+    /**
+     * Identifies the custom event that gets logged when a user action redirects them to a URL set in the effective query pipeline on the search page.
+     */
+    triggerRedirect = 'redirect',
+}
+
+export const CustomEventsTypes: Partial<Record<SearchPageEvents, string>> = {
```
samisayegh, line 78: Similar as above, is the `Partial` keyword needed here?

src/searchPage/searchPageEvents.ts
```diff
@@ -57,6 +57,29 @@ export enum SearchPageEvents {
...
+    [SearchPageEvents.triggerNotify]: 'queryPipelineTriggers',
+    [SearchPageEvents.triggerExecute]: 'queryPipelineTriggers',
+    [SearchPageEvents.triggerQuery]: 'queryPipelineTriggers',
+    [SearchPageEvents.triggerRedirect]: 'queryPipelineTriggers'
```
samisayegh, line 82: Is the value `queryPipelineTriggers` used elsewhere in the project? If so, it would be better to store as a `const` and reference.

src/client/analytics.ts
olamothe: By changing it to a `Record<string, string>`, nope !

src/searchPage/searchPageEvents.ts
olamothe: Yes, for this one we need a partial, otherwise we need to specify each `SearchPageEvents` keys in that dictionary, but we only want to have that info for "customEvents".

src/searchPage/searchPageEvents.ts
olamothe: It's not used anywhere else, no.

PR #200, opened 2020-05-11
Initial work to support running in node js environment

olamothe:
For headless, one requirement is that we need to be able to run the library on a browser as well as inside a node js environment.

This PR just adds various checks around availability of globals that would otherwise throw exception in a node js env. 

This means the script should now be importable in node js, but it does not actually make it "usable" and working properly. The main culprit for now is the "CookieStorage", for which I will open a subsequent PR.

For now, I've set it up as a "NullStorage", with a console warning.


src/client/analytics.ts
```diff
@@ -32,10 +32,13 @@ export const Endpoints = {
     hipaa: 'https://usageanalyticshipaa.coveo.com',
 };
 
+export type SupportedClientEnvironment = 'node' | 'browser'
```
dlafreniere, line 35: would there ever be any other ones? can we reduce this to simply `isBrowser`? and then re-use that instead of `this.options.env == 'browser'`

src/client/analytics.ts
```diff
@@ -107,17 +112,27 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        this.analyticsBeaconClient = this.options.env == 'browser' ? new AnalyticsBeaconClient(clientsOptions) : null;
         this.analyticsFetchClient = new AnalyticsFetchClient(clientsOptions);
-        window.addEventListener('beforeunload', () => this.flushBufferWithBeacon());
+        this.options.env == 'browser' && hasWindow() ? window.addEventListener('beforeunload', () => this.flushBufferWithBeacon()) : null;
+    }
+
+    private initStorage(): WebStorage {
+
+        if (this.options.env === 'node') {
+            console.log('Storage not implemented for nodejs environments')
```
dlafreniere, line 123: (nitpick): `Node.js`

src/detector.ts
```diff
@@ -1,19 +1,35 @@
...
 export function hasCookieStorage(): boolean {
-    return navigator.cookieEnabled;
+    return hasNavigator() ? navigator.cookieEnabled : false
 }
+
+export function hasCrypto(): boolean {
+    return typeof crypto !== 'undefined'
+}
+
+export function hasCryptoRandomValues(): boolean {
```
dlafreniere, line 29: I don't think you can have crypto without getRandomValues (not sure if its worth having separate detector functions)

src/client/analytics.ts
olamothe: It's not 100% clear for now, but eventually we will want to support stuff other than "node", yes. 

For example, for commerce, we will want to support mobile native app.

There are way to import JS library using NativeScript/ReactNative/JavaScriptCore etc..

For those, it is similar to node env, but with some "twist" :) 

When we did some POC for headless, we saw that very often, node js lib could be imported directly without any changes using ReactNative and NativeScript. JavaScript core was a different beast altogether.

It might be possible that we will need to adapt some libraries we use, or code some "shim" that would not be needed for browser or node.

But, since I don't really know what (if anything) we will actually need to change, I guess we could also simplify and just do `isBrowser`.

What do you prefer ?

src/client/analytics.ts
```diff
@@ -127,12 +142,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     set currentVisitorId(visitorId: string) {
         this.visitorId = visitorId;
-        if (hasCookieStorage()) {
-            this.cookieStorage.setItem('visitorId', visitorId);
-        }
-        if (hasLocalStorage()) {
-            localStorage.setItem('visitorId', visitorId);
-        }
+        this.storage.setItem('visitorId', visitorId);
```
olamothe, line 145: This is one thing I wasn't too sure. It seems the old code was using both local storage and cookies to store info.

Was that intentional ? Should it be brought back ?

src/client/analytics.ts
dlafreniere: that sounds legit, I just found `this.options.env == 'browser'` and `this.options.env == 'node'` a bit... verbose?

olamothe: I could add a getter :

```
private get isBrowser() {
 return this.options.env === 'browser';
}
```

src/client/analytics.ts
```diff
@@ -107,17 +112,27 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             token,
             visitorIdProvider: this,
         };
-        this.analyticsBeaconClient = new AnalyticsBeaconClient(clientsOptions);
+        this.analyticsBeaconClient = this.options.env == 'browser' ? new AnalyticsBeaconClient(clientsOptions) : null;
         this.analyticsFetchClient = new AnalyticsFetchClient(clientsOptions);
-        window.addEventListener('beforeunload', () => this.flushBufferWithBeacon());
+        this.options.env == 'browser' && hasWindow() ? window.addEventListener('beforeunload', () => this.flushBufferWithBeacon()) : null;
```
francoislg, line 117: This is more of a general comment (not just for this line), but would it be possible to encapsulate all the "browser" logic into its own class?

I could see something like...

```
private environment: EnvironmentSpecificClient;

this.environment = isBrowser() ? new BrowserEnvironment(this) : new NodeJSEnvironment();
```

it would make it very simple to split some implementations depending on the environment, and it would remove *most* of the `if`s from this PR, such as:

```
this.storage = this.environment.createStorage();
```

we could even encapsulate the `window.beforeUnload` into a `this.environment.init()` or something like that.

Thoughts?

src/client/analytics.ts
francoislg: I was intentional. If a user sets a new visitor ID through `coveoua("set", "userId", "{newvalue}")`, the change must be propagated in the Cookies so that the JSUI also uses this visitor ID

src/client/analytics.ts
samisayegh: I find the name `RuntimeEnvironment` better captures the purpose of the property.

src/client/analytics.ts
```diff
@@ -107,17 +112,27 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             token,
             visitorIdProvider: this,
         };
-        this.analyticsBeaconClient = new AnalyticsBeaconClient(clientsOptions);
+        this.analyticsBeaconClient = this.options.env == 'browser' ? new AnalyticsBeaconClient(clientsOptions) : null;
```
samisayegh, line 115: Would it be safer to return a no-op `NullAnalyticsBeaconClient` here respecting the beacon client interface?

src/client/analytics.ts
samisayegh: I support this suggestion. If we have a third environment, it will be much easier to extend if we have multiple implementations respecting one interface.

src/client/analytics.ts
```diff
@@ -194,7 +204,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
     private flushBufferWithBeacon(): void {
         while (this.hasPendingRequests()) {
             const {eventType, payload} = this.bufferedRequests.pop() as BufferedRequest;
-            this.analyticsBeaconClient.sendEvent(eventType, payload);
+            this.analyticsBeaconClient?.sendEvent(eventType, payload);
```
samisayegh, line 207: If we used a null client instead of initializing to `null`, we would avoid any optional chaining gotchas.

rollup.config.js
```diff
@@ -31,8 +30,20 @@ const umd = {
...
 }
 
+const libUMD = {
+    input: './src/coveoua/library.ts',
+    output: [
+        umdConfig('./dist/library.js'),
+    ],
+    plugins: [
+        tsPlugin(),
+        uglify()
```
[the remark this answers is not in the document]
olamothe: You can import that common js lib in a browser, though (because you cannot use es modules for any reason). 

That's what we do in JSUI currently.

So the not "sensitive to the payload size" is not totally accurate.

src/client/analytics.ts
olamothe: Noted !

src/client/analytics.ts
olamothe: Yes good idea !

src/client/analytics.ts
olamothe: Alright !

src/client/analytics.ts
olamothe: Will revert then.

olamothe:
Closing PR: Will re-open with environment class suggestion

PR #201, opened 2020-06-05
Re-opened PR to support NodeJS runtime

olamothe:
I closed my old PR (got sidetracked on a ton of different stuff since that PR ;) )

It's similar to the previous one I had (we need to be able to support Node JS environment for coveo headless), except I implement the "Runtime" interface that was suggested by @francoislg .


src/detector.ts
```diff
@@ -1,19 +1,35 @@
...
-        return 'sessionStorage' in window && window['sessionStorage'] !== null;
-    } catch (e) {
-        return false;
-    }
+    return typeof sessionStorage !== 'undefined';
 }
 
 export function hasCookieStorage(): boolean {
-    return navigator.cookieEnabled;
+    return hasNavigator() ? navigator.cookieEnabled : false
```
francoislg, line 22:
```suggestion
    return hasNavigator() && navigator.cookieEnabled;
```

src/storage.ts
```diff
@@ -38,6 +38,24 @@ export class CookieStorage implements WebStorage {
     }
 }
 
+export class CookieAndLocalStorage implements WebStorage {
```
francoislg, line 41: Really cool idea, I like that!

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.1.0",
+    "version": "2.1.1",
```
dlafreniere, line 3: not a minor?

PR #203, opened 2020-06-09
Change endpoints to platform proxy

francoislg:
[COM-470]

[COM-470]: https://coveord.atlassian.net/browse/COM-470

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -16,7 +16,7 @@ import {
     IRequestPayload,
     VariableArgumentsPayload,
 } from '../events';
-import {VisitorIdProvider, AnalyticsRequestClient} from './analyticsRequestClient';
+import {VisitorIdProvider} from './analyticsRequestClient';
 import {hasWindow, hasDocument} from '../detector';
 import {addDefaultValues} from '../hook/addDefaultValues';
 import {enhanceViewEvent} from '../hook/enhanceViewEvent';
@@ -27,9 +27,9 @@ import {IRuntimeEnvironment, BrowserRuntime, NodeJSRuntime} from './runtimeEnvir
 export const Version = 'v15';
 
 export const Endpoints = {
-    default: 'https://usageanalytics.coveo.com',
-    production: 'https://usageanalytics.coveo.com',
-    hipaa: 'https://usageanalyticshipaa.coveo.com',
+    default: 'https://platform.cloud.coveo.com/rest/ua',
+    production: 'https://platform.cloud.coveo.com/rest/ua',
+    hipaa: 'https://platformhipaa.cloud.coveo.com/rest/ua',
 };
 
 export interface ClientOptions {
@@ -331,7 +331,8 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
 
     private get baseUrl(): string {
         const {version, endpoint} = this.options;
-        return `${endpoint}/rest/${version}`;
+        const endpointIsCoveoProxy = endpoint.indexOf(".cloud.coveo.com") !== -1;
+        return `${endpoint}${endpointIsCoveoProxy ? '' : '/rest'}/${version}`;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/client/analytics.ts
```diff
@@ -331,7 +331,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     private get baseUrl(): string {
         const {version, endpoint} = this.options;
-        return `${endpoint}/rest/${version}`;
+        return `${endpoint}/${version}`;
```
francoislg, line 334: hm...

This is kind of a breaking change.

Say I am passing a custom endpoint, now I am removing `/rest` from the path...

I have no idea how to cleanly fix this though... Any idea?

toussaints: Well, are custom endpoint even supported?

francoislg: Yeah, this is the third parameter of the init, `coveoua("init", "token", "endpoint | client")`.

It is also used in Coveo for Sitecore with `coveo("initForProxy", "endpoint")`

src/client/analytics.ts
```diff
@@ -331,7 +331,8 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     private get baseUrl(): string {
         const {version, endpoint} = this.options;
-        return `${endpoint}/rest/${version}`;
+        const alreadyHasRestInPath = endpoint.indexOf("/rest") !== -1;
```
toussaints, line 334: maybe we could compare with default endpoints and leave it as-is when it's a value that is not in the const?

francoislg: Yeah... but I feel like adding the `/rest` was a bad idea in the first place...

And it will prevent us from using other environments with the proxy endpoint, like dev and qa 😖

francoislg: ok so what about if the endpoint contains `*.cloud.coveo.com`

francoislg:
Take 3, now it will simply not append `/rest` if the path contains `.cloud.coveo.com`

src/client/analytics.ts
```diff
@@ -331,7 +331,8 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     private get baseUrl(): string {
         const {version, endpoint} = this.options;
-        return `${endpoint}/rest/${version}`;
+        const endpointIsCoveoProxy = endpoint.indexOf(".cloud.coveo.com") !== -1;
+        return `${endpoint}${endpointIsCoveoProxy ? '' : '/rest'}/${version}`;
```
toussaints, line 335: I think this is the best we can do..

src/client/analytics.ts
```diff
@@ -331,7 +331,8 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     private get baseUrl(): string {
         const {version, endpoint} = this.options;
-        return `${endpoint}/rest/${version}`;
+        const endpointIsCoveoProxy = endpoint.indexOf(".cloud.coveo.com") !== -1;
```
toussaints, line 334: do you have examples of sitecore's endpoints? do they contain cloud.coveo.com?

src/client/analytics.ts
dlafreniere: we _could_ parse the url and check if the domain is really cloud.coveo.com, and even check if /rest is the first part after :trollface: but _given a reasonable worthwhile effort_, this is the best we can do 😇

src/client/analytics.ts
francoislg: Coveo for Sitecore ships with a custom endpoint, `/coveo/rest`, that injects the access token in the headers.

So the domain is actually the client's domain, followed by `/coveo/rest`, and `/coveo/rest/coveoanalytics` redirects to the analytics endpoint.

They don't contain `cloud.coveo.com` 😄

PR #204, opened 2020-06-15
Quick win for payload size optimization for headless

olamothe:
Currently, coveo.analytics stands at about ~25-40% of the payload size for headless (depending on the module type).

In general, the way out of this problem is that we want to do much more analytics recording at the API level, and so remove most of the logic needed in the front end, meaning most of the code in this library will hopefully become obsolete in the end.

This is an ongoing project that will require some time though. 

With these relatively simple change proposed in this PR (exporting only the Search Page client, and targeting es6 only), we're able to reduce the final payload of coveo.analytics by about 30%.

There could be more optimization we could do toimprove some modules in the final payload, but none of the changes are trivial, and require quite a bit of code change in the library.

Here is the breakdown of each module and their relative size in the bundle: 

image

I would release this as version 2.3.0 if/when approved.

TY !


src/coveoua/headless.ts
```diff
@@ -0,0 +1 @@
+export {CoveoSearchPageClient,SearchPageClientProvider} from '../searchPage/searchPageClient'
```
francoislg, line 1: So in NodeJS, we will only have access to `CoveoSearchPageClient` and `SearchPageClientProvider` ? 🤔 

Should it also export `CoveoAnalyticsClient`?

olamothe: Currently, inside headless, we don't need to reference `CoveoAnalyticsClient` directly (all the functionalities we need from headless POV is exposed in `CoveoSearchPageClient`). So this does not seem necessary for now.

`CoveoAnalyticsClient` is included in the final bundle, though (CoveoSearchPageClient inherits/import it).

francoislg: Sorry, I meant to ask: 

Should we expect _everyone_ to use `CoveoSearchPageClient` and nothing else? There might be cases where we would want the low-level client, like if I already have implemented the Search Page state?

olamothe: Sure thing, in the end it does not affect the bundle size, so I did the modification and exported it.

PR #206, opened 2020-06-22
Support analytics disabling

olamothe:
For headless, we need to be able to specify if analytics tracking is enabled or not. By default, we will be tracking all analytics event, but we need to be able to opt out if need be.

As far as Headless is concerned, the code won't be impacted too much since we will always be calling "logThis(), logThat()", which is the reason I've decided to add the "Noop" logic in coveo.analytics instead of Headless itself.

We also need a way to be able to clear any analytics cookies after initialization.

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -21,8 +21,13 @@ import {hasWindow, hasDocument} from '../detector';
 import {addDefaultValues} from '../hook/addDefaultValues';
 import {enhanceViewEvent} from '../hook/enhanceViewEvent';
 import {uuidv4} from './crypto';
-import {convertKeysToMeasurementProtocol, isMeasurementProtocolKey, convertCustomMeasurementProtocolKeys} from './measurementProtocolMapper';
+import {
+    convertKeysToMeasurementProtocol,
+    isMeasurementProtocolKey,
+    convertCustomMeasurementProtocolKeys,
+} from './measurementProtocolMapper';
 import {IRuntimeEnvironment, BrowserRuntime, NodeJSRuntime} from './runtimeEnvironment';
+import HistoryStore from '../history';
 
 export const Version = 'v15';
 
@@ -56,7 +61,7 @@ export interface AnalyticsClient {
     getHealth(): Promise<HealthResponse>;
     registerBeforeSendEventHook(hook: AnalyticsClientSendEventHook): void;
     addEventTypeMapping(eventType: string, eventConfig: EventTypeConfig): void;
-    runtime: IRuntimeEnvironment
+    runtime: IRuntimeEnvironment;
 }
 
 interface BufferedRequest {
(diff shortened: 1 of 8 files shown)


src/client/analytics.ts
```diff
@@ -214,6 +220,12 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return this.bufferedRequests.length > 0;
     }
 
+    public clear() {
+        this.storage.removeItem('visitorId');
+        const store = new HistoryStore();
+        store.clear();
```
samisayegh, line 226: Why do we create a new history store and call `clear` on it? It feels like a work-around to reset global state?
Would it be difficult to pass the store instance to the Analytics client?

dlafreniere:
>As far as Headless is concerned, the code won't be impacted too much since we will always be calling "logThis(), logThat()", which is the reason I've decided to add the "Noop" logic in coveo.analytics instead of Headless itself.

it kinda seems to me like the responsibility of choosing whether to log or not belongs to the client (i.e. Headless in this case) and not the logging lib (coveo.analytics), however from an implementation stand point, it seems to be fairly simple this way

in the `logThis()` and `logThat()` we could have had something like this:
```
function logThis() {
  if (!analyticsDisabled) {
    coveoAnalytics.logThis();
  }
}
```

olamothe:
> it kinda seems to me like the responsibility of choosing whether to log or not belongs to the client (i.e. Headless in this case) 

Which it will be: the configuration and control of "should I log analytics or not" is ultimately headless responsibility.

What this PR propose, instead, is an API to allow the client (headless) to achieve that without having too much knowledge about what needs to be done exactly.

For your debugging example, IMHO a better API, if I had to choose between two logging library, would be one where I'm able to control the debug level however I see fit, and remove all the if peppered around in my own code.

eg:

```
function initLoggingLibrary() {
  logger.disable()
  // OR
  logger.enable()
  logger.setLevel(DEBUG)
 etc...
}

function somewhereInMyCode() {
   // this seems easier/nicer to use as I can simply remove all the if everywhere.
   logger.debug("foo")
}

```

Also, this means the consumer library has to know that it should clear some cookies when analytics is disabled. Meaning, we'd still need to call some kind of clear() function (which IMHO is not self evident from the library consumer point of view).

Thoughts ?

src/client/analytics.ts
olamothe: Yeah, to achieve that I'd potentially have to refactor quite a bit of code it seems. 

And yes, it's a "global state" in the sense that an `HistoryStore` is pretty much just `an object that knows which key to set/get/delete in local storage`, and local storage is global :D 

I can look at what it would take to remove the global store, if it gets silly/too much change I'll  go with what I have right now.

olamothe: So, yeah, currently the store is being created in a hook for `enhanceViewEvent`, I could create a reference to the store inside analytics.ts itself, but it seems weird and not really worth it in the end...

ThibodeauJF:
I think it's just not headless that will benefit from a feature like this, but any consumers of that library. For GDPR compliance for example, I may want to start with an already setup but disabled Analytics library, and enable it without too much hassle as soon as an end users accepts "being tracked", or the other way around by clearing history.
With that in mind I agree it's a necessary feature for an analytics library.

dlafreniere:
Sounds good to me 👍

PR #207, opened 2020-06-29
Fix formatting issue when sending beacon with collect

francoislg:
[COM-495]

The beacon format was correct for legacy events (all the payload under a `customEvent` key), but with the measurement protocol, each key-value should be sent directly in the payload.

It fixes that, plus introduces tests to properly validate that

[COM-495]: https://coveord.atlassian.net/browse/COM-495

### DIFF
--- a/src/client/analyticsBeaconClient.spec.ts
+++ b/src/client/analyticsBeaconClient.spec.ts
@@ -2,11 +2,26 @@ import {AnalyticsBeaconClient} from './analyticsBeaconClient';
 import {EventType} from '../events';
 
 describe('AnalyticsBeaconClient', () => {
+    const baseUrl = 'https://bloup.com';
+    const token = '👛';
+    const currentVisitorId = 'hereiam';
+
+    const originalSendBeacon = navigator.sendBeacon;
+    const sendBeaconMock = jest.fn();
+    beforeAll(() => {
+        navigator.sendBeacon = sendBeaconMock;
+    });
+
+    afterAll(() => {
+        navigator.sendBeacon = originalSendBeacon;
+    })
+
+    beforeEach(() => {
+        jest.clearAllMocks();
+    })
+
+
     it('can send an event', async () => {
-        navigator.sendBeacon = jest.fn();
-        const baseUrl = 'https://bloup.com';
-        const token = '👛';
-        const currentVisitorId = 'hereiam';
         const eventType: EventType = EventType.custom;
 
         const client = new AnalyticsBeaconClient({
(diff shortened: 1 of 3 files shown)


src/client/analyticsBeaconClient.spec.ts
```diff
@@ -2,11 +2,26 @@ import {AnalyticsBeaconClient} from './analyticsBeaconClient';
...
+    const currentVisitorId = 'hereiam';
+
+    const originalSendBeacon = navigator.sendBeacon;
+    const sendBeaconMock = jest.fn();
+    beforeAll(() => {
+        navigator.sendBeacon = sendBeaconMock;
+    });
+
+    afterAll(() => {
+        navigator.sendBeacon = originalSendBeacon;
```
toussaints, line 16: why do we have to set and unset the original beacon client?

src/client/analyticsBeaconClient.spec.ts
```diff
@@ -21,6 +36,60 @@ describe('AnalyticsBeaconClient', () => {
...
+        });
+
+        expect(sendBeaconMock).toHaveBeenCalledWith(`${baseUrl}/analytics/collect?access_token=👛&visitorId=${currentVisitorId}`, jasmine.anything());
+        expect(await getSendBeaconFirstCallBlobArgument()).toBe("pr1a=value&to%20encode=to%20encode");
+    });
+
+    it('can send a collect event with a more complex payload', async () => {
+        const eventType: EventType = EventType.collect;
+
+        const client = new AnalyticsBeaconClient({
```
toussaints, line 66: we create a beacon client in a similar fashion in all those test methods couldn't we put that in the beforeEach?

src/client/analyticsBeaconClient.spec.ts
```diff
@@ -21,6 +36,60 @@ describe('AnalyticsBeaconClient', () => {
...
+
+    const getSendBeaconFirstCallBlobArgument = async () => {
+        const blobArgument: Blob = sendBeaconMock.mock.calls[0][1];
+        expect(blobArgument.size).toBeGreaterThan(0);
+        return new Promise((resolve) => {
+            const reader = new FileReader();
+            reader.addEventListener('loadend', () => {
+                resolve(reader.result);
+            });
+            reader.readAsText(blobArgument);
```
toussaints, line 92: Woah that is some black magic

francoislg: Usually, we can use `await blob.text()`, but it was not accessible and I don't know why :/ 

This is me giving up, and finding a way to just-make-it-work™️

src/client/analyticsBeaconClient.spec.ts
francoislg: It is modifying a global variable, I wanted to be sure we were not messing up the implementation somewhere else

src/client/analyticsBeaconClient.ts
```diff
@@ -37,13 +32,50 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
...
+                  access_token: this.opts.token,
+                  ...payload,
+              });
+    }
+
+    private getQueryParamsForEventType(eventType: EventType): string {
+        const {token, visitorIdProvider} = this.opts;
+        const visitorId = visitorIdProvider.currentVisitorId;
+        return [
+            token && this.isEventTypeLegacy(eventType) ? `access_token=${token}` : '',
```
francoislg, line 48: @bigluck this fixed the issue you mentioned (plus line 39)

bigluck: Top!

PR #210, opened 2020-06-30
[COM-400] set userId to anonymous for API keys

dlafreniere:
When UA receives events without a userId, it adds one based on the token UNLESS the event is anonymous. However, the measurement protocol does not include a field for UA to identify if the event is supposed to be anonymous or not. Therefore, to prevent adding an 'API Key' user, we explicitly set the userId to `anonymous` which gets picked up properly from UA

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -45,9 +45,9 @@ describe('ec events', () => {
     });
 
     it('can set custom data at the root level and send a page view event', async () => {
-        await coveoua("set", "custom", {
-            "verycustom": "value"
-        })
+        await coveoua('set', 'custom', {
+            verycustom: 'value',
+        });
         await coveoua('send', 'pageview');
 
         const [body] = getParsedBody();
@@ -78,7 +78,7 @@ describe('ec events', () => {
 
     it('can send a product detail view event with custom values', async () => {
         coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand'});
-        coveoua('ec:setAction', 'detail', {storeid: 'amazing', custom: { verycustom: "value"}});
+        coveoua('ec:setAction', 'detail', {storeid: 'amazing', custom: {verycustom: 'value'}});
         await coveoua('send', 'pageview');
 
         const [body] = getParsedBody();
@@ -90,7 +90,7 @@ describe('ec events', () => {
             pr1id: 'something',
             pr1br: 'brand',
             pa: 'detail',
-            verycustom: "value",
+            verycustom: 'value',
         });
     });
 
@@ -116,8 +116,8 @@ describe('ec events', () => {
             title: 'wow',
             location: 'http://right.here',
             custom: {
-                verycustom: "value"
-            }
+                verycustom: 'value',
+            },
         });
 
         const [body] = getParsedBody();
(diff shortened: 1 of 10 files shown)


functional/ec-events.spec.ts
```diff
@@ -207,6 +207,71 @@ describe('ec events', () => {
...
+
+            it('should not set the user id', async () => {
+                await coveoua('send', 'pageview');
+
+                const [event] = getParsedBody();
+
+                expect(Object.keys(event)).not.toContain('uid');
+            });
+            it('should not overwrite existing userId', async () => {
+                const steveJobs = 'steve@apple.com';
```
cnguyen-coveo, line 260: 🍎

src/client/token.ts
```diff
@@ -0,0 +1,3 @@
+const API_KEY_PREFIX = 'xx';
+
+export const isApiKey = (token?: string) => token?.startsWith(API_KEY_PREFIX);
```
francoislg, line 3: Nitpick, but maybe add `|| false` so that the type is simply a boolean? I think the type is currently `boolean | undefined`

src/client/analytics.ts
```diff
@@ -341,6 +345,14 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return rest;
     }
 
+    private setAnonymousUser(payload: IRequestPayload): IRequestPayload {
+        const {userId, ...rest} = payload;
+        if (!userId) {
```
francoislg, line 350: 🤔 Passing `{ userId: 'someone', a: 'b' }` here will result in `{ a: 'b' }`. It is quite weird because the method is called `setAnonymousUser`, but in fact, it either removes the userId or set the anonymous one, which makes sense *only* if the user is using an API key.

So I think you should bring back the `if (isApiKey())` here for clarity, or change the method name to something like `ensureAnonymousUserWhenUsingApiKey` 😄

functional/ec-events.spec.ts
```diff
@@ -207,6 +207,71 @@ describe('ec events', () => {
...
+            });
+
+            it('should not set the user id', async () => {
+                await coveoua('send', 'pageview');
+
+                const [event] = getParsedBody();
+
+                expect(Object.keys(event)).not.toContain('uid');
+            });
+            it('should not overwrite existing userId', async () => {
```
samisayegh, line 259: In this case, where both a `token` and `userId` are defined, it seems that intent is a non-anonymous UA event. In such a situation, should we use the token as the source of truth?

Said differently, could the current mechanism be abused if an authenticated user wants to make their search activity appear like it was performed by someone else?

src/client/analytics.ts
dlafreniere: sounds good,  I wasn't quite decided on where that check should fall. I saw a couple of the other steps had "preconditions" in the actual step definition

functional/ec-events.spec.ts
dlafreniere: The behaviour is actually managed by UA:
> username (string, optional): The name of the user security identity who caused the search interface to log the event. Can only be specified when authenticated with a token with impersonate privilege. If unspecified and the request is authenticated with a search token, the service attempts to extract the username value from the access token that authenticated the request. If this fails, the username value is left blank. If unspecified and the request is authenticated with an API key, username is set to platform@coveo.com

basically, we want to set the userId explicitly to anonymous for the case of and API key since the measurement protocol doesn't have a field for isAnonymous.

PR #213, opened 2020-07-28
Add support for originLevel in search page client

olamothe:
originLevel1/2/3 support.

Also, I made them mandatory, even though they are in theory not required by the API.

In practice though, you'd want to always set all parameters for proper support in Coveo/being able to support all types of advanced ML models/build all types of reports etc.

More data > less data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.5.0",
+    "version": "2.6.0",
     "description": "📈 Coveo analytics client (node and browser compatible) ",
     "main": "dist/library.js",
     "module": "dist/library.es.js",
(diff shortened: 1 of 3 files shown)


dlafreniere:
> originLevel1/2/3 support.
> 
> Also, I made them mandatory, even though they are in theory not required by the API.
> 
> In practice though, you'd want to always set all parameters for proper support in Coveo/being able to support all types of advanced ML models/build all types of reports etc.
> 
> More data > less data.

Completely agree, but are you sure that most orgs use all 3? I vaguely remember something from UA, where most orgs have level 1, but not always 2 and 3

olamothe:
> Completely agree, but are you sure that most orgs use all 3? I vaguely remember something from UA, where most orgs have level 1, but not always 2 and 3

Yeah, which has been a problem in the past. Most org *should*, but they don't.

PR #214, opened 2020-07-30
[COM-395] Warn of a too small position when one is provided

Spuffynism:
Log a warning message for product or impression positions smaller than 0, but still add them.

### DIFF
--- a/src/plugins/ec.ts
+++ b/src/plugins/ec.ts
@@ -136,20 +136,8 @@ export class EC {
             ...(this.actionData || {}),
         };
 
-        const productPayload = this.products.reduce((newPayload, product, index) => {
-            return {
-                ...newPayload,
-                ...convertProductToMeasurementProtocol(product, index),
-            };
-        }, {});
-
-        const impressionsByList = this.getImpressionsByList();
-        const impressionPayload = impressionsByList.reduce((newPayload, impressionList, index) => {
-            return {
-                ...newPayload,
-                ...convertImpressionListToMeasurementProtocol(impressionList, index),
-            };
-        }, {});
+        const productPayload = this.getProductPayload();
+        const impressionPayload = this.getImpressionPayload();
 
         this.clearData();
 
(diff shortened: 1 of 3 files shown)


src/plugins/ec.ts
```diff
@@ -85,10 +85,18 @@ export class EC {
     }
 
     addProduct(product: Product) {
+        if (product.position !== undefined && product.position < 1) {
```
francoislg, line 88: :thinking: I'm skeptical about adding one validation _here_ specifically because we have managed to avoid it by now.

Like, it could easily be validated in the `addECDataToPayload` process:


```
const productPayload = this.products.reduce((newPayload, product, index) => {
            return {
                ...newPayload,
                ...convertProductToMeasurementProtocol(product, index),
            };
        }, {});
```

:arrow_down: 


```
const productPayload = this.products.map(product => this.cleanUpProduct(product)).reduce((newPayload, product, index) => {
            return {
                ...newPayload,
                ...convertProductToMeasurementProtocol(product, index),
            };
}, {});

...

private cleanUpProduct(product: Product) {
   const {position, ...productRest} = product;
   if (product.position < 1) {
        console.warn("...");
        return productRest;
   } else {
        return product;
  }
}
```

It would match a little more what we have in the analytics.ts part: https://github.com/coveo/coveo.analytics.js/blob/7ef4ead28771d51524991ec56842f7b6425a7e7c/src/client/analytics.ts#L338 

Thoughts?

francoislg:
As a side note (reflected in my comment), I think we should outright ignore invalid values since the API would break anyway

src/plugins/ec.ts
Spuffynism: Hmmm, I felt like placing the warning in the `addProduct` and `addImpression` methods was more appropriate because I was thinking that we'd like to know that a product (or an impression) with an invalid position was created **as soon as possible**. Thoughts on this?

If I were to add the validation in a `cleanUpProduct` step, I'd also add the name or id of the product/impression that has an invalid position. And since we're only validating it and not changing it I'd do something more like this:
```
private assureProductValidity(product: Product) {
    if (product.position !== undefined && product.position < 1) {
        console.warn(`The position for product '{product.name || product.id}' must be greater than 0 when provided.`);
    }

    return product;
}
```
What do you think?

I'll trust your judgment on where's the best place to add this validation. Avoiding the validation in `addProduct` or `addImpression` sure would be nice!

Spuffynism:
@francoislg I applied your suggestions. Let me know what you think!

francoislg:
@Spuffynism looks good, the only thing is that I am still not sure we should keep the value. The back-end will eventually throw for this :/ 

Let's say we ignore position if invalid, worst case is that it will keep its default value, so I think it does not outweigh losing the rest of the data

Spuffynism:
@francoislg Yeah, I understand. Should we close the PR for now and re-open it when/if we decide to warn?

francoislg:
@Spuffynism I still think the warning is nice, we should just modify the product to *remove* `position` from the payload if invalid :P 

This is what I had in my code snippet:

```
  const {position, ...productRest} = product;
   if (product.position < 1) {
        console.warn("...");
        return productRest;
   } else {
        return product;
  }
```

Spuffynism:
@francoislg Haaa, ok I get it now. I agree! I'll make the change. :smile:

src/plugins/ec.ts
dlafreniere: I kinda agree with Nicholas that the validation should occur as soon as possible...

The way it's implemented is good/clean too though

PR #218, opened 2020-10-08
Added getVisitorId() to AnalyticsClient interface

jlblais:
Adding this getter function to the interface in order to use it in headless to send the `visitorId` to the searchAPI.

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -63,7 +63,7 @@ export interface AnalyticsClient {
     registerBeforeSendEventHook(hook: AnalyticsClientSendEventHook): void;
     addEventTypeMapping(eventType: string, eventConfig: EventTypeConfig): void;
     runtime: IRuntimeEnvironment;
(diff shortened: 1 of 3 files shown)


src/client/analytics.ts
```diff
@@ -264,6 +265,10 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     addEventTypeMapping(eventType: string, eventConfig: EventTypeConfig): void {
         this.eventTypeMapping[eventType] = eventConfig;
+    } 
+    
+    getVisitorId(){
```
francoislg, line 270: There already is a `get currentVisitorId` :thinking: 

could it be put in the interface instead?

jlblais: Yes, good point. I added it in readonly mode.

francoislg (approved):
:+1: LGTM

cnguyen-coveo (approved):

PR #221, opened 2020-10-22
Request no response body on sendBeacon to prevent corb warning

Spuffynism:
[COM-665]
Add a GET parameter to `sendBeacon` that will request discarding of the response body and make the response status code a 204 No Content. 

This prevents the CORB warning by leaving no content for the browser to read and therefore to block.

Before:
image

After:
image

[COM-665]: https://coveord.atlassian.net/browse/COM-665

### DIFF
--- a/src/client/analyticsBeaconClient.ts
+++ b/src/client/analyticsBeaconClient.ts
@@ -47,6 +47,7 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
         return [
             token && this.isEventTypeLegacy(eventType) ? `access_token=${token}` : '',
             visitorId ? `visitorId=${visitorId}` : '',
+            'discardVisitInfo=true',
         ]
             .filter((p) => !!p)
             .join('&');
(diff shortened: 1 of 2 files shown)


src/client/analyticsBeaconClient.ts
```diff
@@ -47,6 +47,7 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
         return [
             token && this.isEventTypeLegacy(eventType) ? `access_token=${token}` : '',
             visitorId ? `visitorId=${visitorId}` : '',
+            'discardVisitInfo=true',
```
toussaints, line 50: indent seems odd here

cnguyen-coveo (approved):

toussaints (approved):

src/client/analyticsBeaconClient.ts
Spuffynism: Uhh, everything is aligned, am I missing something? :dizzy_face:

francoislg: I would check with the UA team if the name is fine, so far I think this is the easiest solution :P

dlafreniere (approved):

src/client/analyticsBeaconClient.ts
agagnon-coveo: > indent seems odd here

I think it's because of the single quote being small and centered. It does look odd.

PR #223, opened 2020-11-02
Added utility method to format custom data from array

nrawji:


src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
+function filterRepeatedValues(rawData: string[]) {
```
nrawji, line 1: I wasn't sure where to put this so I created this utils folder but let me know if there is a better location for it.

francoislg:
:thinking: really not sure about that.

I would first like to have some more context about why this was done, I don't fully understand its purpose just from the code.

Also, it seems completely unused in this project, so I am not convinced that it belongs here.

nrawji (commented):
> 🤔 really not sure about that.
> 
> I would first like to have some more context about why this was done, I don't fully understand its purpose just from the code.
> 
> Also, it seems completely unused in this project, so I am not convinced that it belongs here.

Hmm, I think it does make sense for the analytics package to provide some utility methods than can be used across projects to do common functionality. I'd imagine converting an array to a correctly formatted string representation would be a common use case.

src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
...
+    return rawData.map((value) => {
+        return value.replace(/;/g, '');
+    });
+}
+
+function getDataString(data: string[]): string {
+    const analyticsLengthLimit = 256;
+
+    const formattedData = data.join(';');
+    if ( formattedData.length <= analyticsLengthLimit) {
```
francoislg, line 20: I think you are missing a formatter pass, this space should be removed

src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
...
+        return formattedData;
+    }
+    return getDataString(data.slice(1));
+}
+
+export const formattedCustomDataFromArray = (rawData: string[]) => {
```
francoislg, line 26: I think there is some value to add `: string` as a return value since this is the only exported method, and we want to ensure that it always returns a string (I think)

francoislg, line 26: Nitpick, but I think `formatArrayForCoveoCustomData` would make its purpose a bit clearer?

src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
+function filterRepeatedValues(rawData: string[]) {
+    let prev = '';
+    return rawData.filter((value) => {
```
francoislg, line 3: Looking at this code, `[1,2,1]` will not be filtered? So maybe we should name the method `filterConsecutiveRepeatedValues`?

src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
...
+    const formattedData = data.join(';');
+    if ( formattedData.length <= analyticsLengthLimit) {
+        return formattedData;
+    }
+    return getDataString(data.slice(1));
+}
+
+export const formattedCustomDataFromArray = (rawData: string[]) => {
+    const dataWithoutRepeatedValues = filterRepeatedValues(rawData);
+    const dataWithoutSemicolons = removeSemicolons(dataWithoutRepeatedValues);
```
francoislg, line 28: should the semi-colons be removed _before_ the repeated values? For instance, should `["1;", "1", "2"] have its first `1` filtered out?

src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+function getDataString(data: string[]): string {
+    const analyticsLengthLimit = 256;
+
+    const formattedData = data.join(';');
+    if ( formattedData.length <= analyticsLengthLimit) {
+        return formattedData;
+    }
+    return getDataString(data.slice(1));
```
francoislg, line 23: Nitpick, but you can probably do this non-recursively with a simple while, something like:

```
let formattedData = data.join(';');
let index = 0;
while (formattedData.length < analyticsLengthLimit) {
  index++
  formattedData = data.slice(index).join(";");
}
return formattedData;
```

src/utils/formatted-custom-data-from-array.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+function removeSemicolons(rawData: string[]) {
+    return rawData.map((value) => {
+        return value.replace(/;/g, '');
+    });
+}
+
+function getDataString(data: string[]): string {
+    const analyticsLengthLimit = 256;
```
francoislg, line 17: Nitpick, this is a real "constant value" so you could put it in uppercase `ANALYTICS_LENGTH_LIMIT`

francoislg (commented):
Still not 100% convinced that this should be exposed like that.

If we can do this formatting automatically, without any user noticing, that would be the best thing.

If we really can't do that, we should instead clearly document that the user *must* call this method to convert array values.

src/utils/formatted-custom-data-from-array.ts
nrawji: Personally I find the recursion easier to understand

src/searchPage/searchPageEvents.ts
```diff
@@ -201,8 +201,8 @@ export interface CategoryFacetMetadata {
 
 export interface OmniboxSuggestionsMetadata {
     suggestionRanking: number;
-    partialQueries: string;
-    suggestions: string;
+    partialQueries: string[];
```
francoislg, line 204: This is a breaking change, maybe it should be `string | string[]` for now? :confused:

olamothe:
I think these function could have been scoped/not exported/internal for `SearchPageClient.ts`

I guess as long as they are not exported to the global scope/don't leak from the `formatting` modules, the end result is relatively similar, although it's dubious that they would be used anywhere else in the repo.

As for the background as to why this is needed, that's what is currently done in JSUI to "sanitize" the analytics payload before sending it, so that it can be successfully used by ML model to improve query suggestions models.

When re-implementing the query suggestion feature in Headless, we thought it would make more sense to have this sanitization code to reside in coveo.analytics directly.

Since that's a goal we've set ourselves with Headless, to have all analytics be as much as a blackbox as possible/have a single source of truth (this repo) for everything analytics related.

olamothe (approved):

nrawji:
> I think these function could have been scoped/not exported/internal for `SearchPageClient.ts`
> 
> I guess as long as they are not exported to the global scope/don't leak from the `formatting` modules, the end result is relatively similar, although it's dubious that they would be used anywhere else in the repo.

I thought it would be cleaner to keep it separate from SearchPageClient since as of right now that file has a pretty standard structure and if we had all the formatting inside that file it would get extremely bloated.

ThibodeauJF (approved):

PR #225, opened 2020-11-10
Added #fetchMoreResults analytics event

btaillon:
Needed for [KIT-170](https://coveord.atlassian.net/browse/KIT-170).

### DIFF
--- a/src/searchPage/searchPageEvents.ts
+++ b/src/searchPage/searchPageEvents.ts
@@ -157,6 +157,10 @@ export enum SearchPageEvents {
      * Identifies the search event that gets logged when a user action (that is not a query) reloads the Recommendations panel with new recommendations.
      */
     recommendationInterfaceLoad = 'recommendationInterfaceLoad',
+    /**
+     * Identifies the search event that gets logged when a user requests more results to be added to the existing result list.
+     */
+    fetchMoreResults = 'fetchMoreResults'
 }
 
 export const CustomEventsTypes: Partial<Record<SearchPageEvents, string>> = {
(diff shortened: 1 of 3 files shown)


dlafreniere (approved):

src/searchPage/searchPageEvents.ts
```diff
@@ -157,6 +157,10 @@ export enum SearchPageEvents {
      * Identifies the search event that gets logged when a user action (that is not a query) reloads the Recommendations panel with new recommendations.
      */
     recommendationInterfaceLoad = 'recommendationInterfaceLoad',
+    /**
+     * Identifies the search event that gets logged when a user requests more results to be added to the existing result list.
+     */
+    fetchMoreResults = 'fetchMoreResults'
```
dlafreniere, line 163: where does this identifier come from?

[the docs](https://docs.coveo.com/en/417/javascript-search-framework/javascript-search-framework-events) only mention `fetchMoreSuccess`

btaillon: It's made up, I didn't know whether we are supposed to create new events this way or use existing ones. The current alternative would be to use `pagerScrolling` and specify a `type: "getMoreResults"` in the metadata.

samisayegh (approved):

ThibodeauJF (approved):

src/searchPage/searchPageEvents.ts
olamothe: It needs to match what currently exists in JSUI, we cannot just invent a new value here.

PR #226, opened 2020-11-12
Finished runtime genericization and react native implementation

nrawji:
For a proof of concept I implemented only the react runtime, this is still a work in progress but the structure should stay the same.

**UPDATE**

Big changes from last commit, I realized that if we continue initializing currentId inside the constructor for `CoveoAnalyticsClient` we run into a race condition where the ID might not be set when sendEvent is called. So I had to refactor the code to make it more async friendly. I a `getCurrentVisitorId` method which acts as a singleton and only initializes the ID when its first called by the `AnalyticsBeaconClient/AnalyticsFetchClient` which were already async.


src/storage.ts
```diff
@@ -1,5 +1,6 @@
 import * as detector from './detector';
 import {Cookie} from './cookieutils';
+import AsyncStorage from "@react-native-async-storage/async-storage";
```
nrawji, line 3: I was trying to defer this import to when the `ReactNativeStorage` is constructed but the problem I ran into was that typescript's import function is async and that causes problems with constructors.

nrawji, line 3: I was trying to defer this import to inside the ReactNativeStorage constructor so that way if you're not using react native you don't need to install the package. The problem I ran into was that Typescript's `import()` function is async and constructors need to be synchronous.

src/client/analytics.ts
```diff
@@ -132,8 +136,9 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return this.runtime.storage;
     }
 
-    private initVisitorId() {
-        const existingVisitorId = this.visitorId || this.storage.getItem('visitorId') || '';
+    private async initVisitorId() {
+        console.log(this.storage.getItem('visitorId'));
```
francoislg, line 140: You should remove this one :eyes:

src/client/analytics.ts
```diff
@@ -104,24 +105,27 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        this.initVisitorId().catch(err => console.error("Could not get a visitor ID, ensure the runtime environment is valid", err));
     }
 
     private initRuntime(clientsOptions: IAnalyticsBeaconClientOptions) {
         if (hasWindow() && hasDocument()) {
             return new BrowserRuntime(clientsOptions, () => this.flushBufferWithBeacon());
-        }
+        } else if (typeof navigator !== 'undefined' && navigator.product == 'ReactNative') {
+            return new ReactNativeRuntime(clientsOptions);
+        }1
```
francoislg, line 126: :thinking: extra `1` here

src/client/analytics.ts
```diff
@@ -104,24 +105,27 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
 
-        return new NodeJSRuntime();
```
francoislg, line 124: From what I see, the `NodeJSRuntime` is now never returned? I think you are missing a condition for it, right?

src/storage.ts
francoislg: So this will be included in all browsers that use our code? :confused: 

Did you check the difference in payload size?

src/client/analytics.ts
```diff
@@ -104,24 +105,27 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
             token: this.options.token,
             visitorIdProvider: this,
         };
 
-        this.runtime = this.initRuntime(clientsOptions);
+        this.runtime = this.options.runtimeEnvironment || this.initRuntime(clientsOptions);
         this.analyticsFetchClient = new AnalyticsFetchClient(clientsOptions);
 
-        this.initVisitorId();
+        this.initVisitorId().catch(err => console.error("Could not get a visitor ID, ensure the runtime environment is valid", err));
```
francoislg, line 118: Not sure about this specific error. What would the client need to do if he ever gets that? Since the runtime is automatically detected, I think it is just most confusing than helping.

Would it be cleaner to just `try/catch` any error in `initVisitorId` and fallback on a valid value if the storage is wrong? And maybe change it to a warning like `Could not get a visitor ID from the current runtime environment storage. A random ID has been generated instead.` (to be refined :smile: )

nrawji: Yeah that's better, I wasn't sure if not being able to store the visitor ID should be treated as an analytics breaking bug.

src/client/analytics.ts
nrawji: Before it was equivalent to `NoopRuntime` so I replaced it with that, that way if we do implement a proper `NodeJSRuntime` environment later it doesn't get used by accident.

nrawji: I refactored this to clean it up a bit and use one NodeJSRuntime for any non-browser environment.

src/storage.ts
nrawji: Since this is a peer dependency only I don't think there should be any significant difference in package size since it's not installed by default. Side note, this is changing with npm 7 where peer dependencies will be installed automatically, so it doesn't appear there will be any way to mark optional dependencies anymore.

I'm just worried that it will cause a runtime error if they don't have it installed. It doesn't appear to right now because the `import` gets compiled into a `require()` which I think just returns undefined if the dependency doesn't exist, so it will only cause a crash if they're in a react native environment (in which case we would want them to install the package anyways.

src/client/analytics.ts
```diff
@@ -132,9 +134,24 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         return this.runtime.storage;
     }
 
-    private initVisitorId() {
-        const existingVisitorId = this.visitorId || this.storage.getItem('visitorId') || '';
-        this.currentVisitorId = existingVisitorId || uuidv4();
+    async getCurrentVisitorId() {
```
nrawji, line 137: Right now there is still the old `get currentVisitorId` and `set currentVisitorId` because I didn't want to introduce any breaking changes by changing the `AnalyticsClient` interface.

nrawji:
The build is failing because I haven't changed `history.ts` and I'm not sure how/if it's possible to refactor history to support an AsyncStorage solution without introducing breaking changes?

src/client/analytics.ts
francoislg: I wonder if we should `@Deprecate` the other getter/setter and use those as the basis?

src/client/analyticsFetchClient.ts
```diff
@@ -47,9 +47,9 @@ export class AnalyticsFetchClient implements AnalyticsRequestClient {
         return [EventType.click, EventType.custom, EventType.search, EventType.view].indexOf(eventType) !== -1;
     }
 
-    private get visitorIdParam() {
+    private async getVisitorIdParam() {
         const {visitorIdProvider} = this.opts;
-        const visitorId = visitorIdProvider.currentVisitorId;
+        const visitorId = await visitorIdProvider.getCurrentVisitorId;
```
francoislg, line 52: :thinking: aren't you missing parenthesis here? `getCurrentVisitorId()`

src/client/analytics.ts
```diff
@@ -132,9 +134,24 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        if (!this.currentVisitorId) {
+            try {
+                const existingVisitorId = this.visitorId || (await this.storage.getItem('visitorId')) || null;
```
samisayegh, line 140: Could we initialize `this.currentVisitorId` in the constructor instead of upon calling `getCurrentVisitorId`?
Also, what is the difference between `this.currentVisitorId` and `this.visitorId`?

src/client/analyticsBeaconClient.spec.ts
```diff
@@ -49,7 +50,8 @@ describe('AnalyticsBeaconClient', () => {
             baseUrl,
             token,
             visitorIdProvider: {
-                currentVisitorId,
+                getCurrentVisitorId: () => {return Promise.resolve(currentVisitorId)},
```
samisayegh, line 53: It's possible to simplify here and on line 30, 79 to `() => Promise.resolve(...)`

package.json
```diff
@@ -19,7 +19,11 @@
         "url": "https://github.com/coveo/coveo.analytics.js.git"
     },
     "license": "MIT",
+    "peerDependencies": {
+        "@react-native-async-storage/async-storage": "^1.13.2"
+    },
     "devDependencies": {
+        "@react-native-async-storage/async-storage": "^1.13.2",
```
samisayegh, line 26: `@react-native-async-storage/async-storage` should be a `dependency` since the code will be used when the coveo.analytics package is running inside react-native.

src/client/analytics.ts
```diff
@@ -110,18 +115,15 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
             visitorIdProvider: this,
         };
 
-        this.runtime = this.initRuntime(clientsOptions);
+        this.runtime = this.options.runtimeEnvironment || this.initRuntime(clientsOptions);
```
samisayegh, line 118: I think it would be better to group and determine the runtime inside `initRuntime` which already receives the full client options object. The function would start by checking the `runtimeEnvironment` property before doing it's usual checks.

src/client/analytics.ts
nrawji: We can't without causing a race condition, constructors have to be synchronous so if we initialize `getCurrentVisitorId` inside the constructor there's a chance it won't be fully initialized by the time the first event is sent. `currentVisitorId` is the old interface that will be marked as deprecated because it only supports synchronous storage implementations, the new functions allow for async storage implementations to be supported as well.

package.json
nrawji: I was trying to avoid this to prevent package size bloat for people who aren't using react-native. With the way its currently implemented, when they initialize the analytics client, it will throw an error saying to install `@react-native-async-storage/async-storage` but if you're not running on react-native and ignore it, the package continues to work normally. What I'm trying to figure out is if there is a way to limit that message to only when the react-native runtime is loaded.

src/client/analytics.ts
olamothe: Open question, but should we do a major version release (3.0) of marking things as `@Deprecate` ?

And then not have to worry about breaking changes ?

What do you guys prefer ?

src/client/runtimeEnvironment.ts
```diff
@@ -28,8 +29,19 @@ export class BrowserRuntime implements IRuntimeEnvironment {
 }
 
 export class NodeJSRuntime implements IRuntimeEnvironment {
-    public storage = new NullStorage();
-    public beaconClient = new NoopAnalyticsBeaconClient();
+    public storage: WebStorage;
+    public beaconClient: AnalyticsFetchClient;
+
+    constructor(beaconOptions: IAnalyticsFetchClientOptions, storage?: WebStorage) {
```
olamothe, line 35: 🤔  Why is there logic for ReactNativeStorage in a NodeJSRuntime ?

Should we instead have a `ReactNativeRuntime`, and then modify `initRuntime` to guess/determine which runtime in `analytics.ts` ?

src/detector.ts
```diff
@@ -33,3 +33,7 @@ export function hasCryptoRandomValues(): boolean {
 export function hasLocation(): boolean {
     return typeof location !== 'undefined';
 }
+
+export function isReactNative(): boolean {
+    return typeof navigator !== 'undefined' && navigator.product == 'ReactNative';
```
olamothe, line 38: Is this the documented/official way to determine this ?

src/storage.ts
```diff
@@ -1,10 +1,11 @@
 import * as detector from './detector';
 import {Cookie} from './cookieutils';
+import AsyncStorage from '@react-native-async-storage/async-storage';
 
 export let preferredStorage: WebStorage | null = null;
 
 export interface WebStorage {
-    getItem(key: string): string | null;
+    getItem(key: string): string | null | Promise<string | null>;
```
olamothe, line 8: If we went the breaking change route (3.0), we could simplify this to always be `Promise<string | null>`.

And for web storage (local storage), we could just wrap the return in a Promise directly

src/storage.ts
```diff
@@ -1,10 +1,11 @@
 import * as detector from './detector';
 import {Cookie} from './cookieutils';
+import AsyncStorage from '@react-native-async-storage/async-storage';
```
olamothe, line 3: Another solution would be to have the user of the library supply the dependency themselves. 

It would complexify the usage of the library for end users, but we could save on the bundle import.

Another approach would be to use rollup code splitting feature : https://rollupjs.org/guide/en/#code-splitting

In any case, I don't think it's acceptable to always load the dependency upfront when 99% of the current users of the library don't require it.

olamothe (commented):
I'd like to know the others opinion about doing a major version release instead of `@Deprecate`.

TY !

src/client/analytics.ts
nrawji: I'd prefer just creating a new major release, I think having deprecated methods that simply won't work in react-native will make maintenance more complicated.

src/detector.ts
nrawji: I didn't find it in the official documentation but there is this github thread https://github.com/facebook/react-native/issues/1331 and this commit https://github.com/facebook/react-native/commit/3c65e62183ce05893be0822da217cb803b121c61

src/client/analytics.ts
samisayegh: For the constructor, that makes senses. For `this.currentVisitorId`, could we keep this property name on the new interface to avoid introducing `this.visitorId`?

src/storage.ts
francoislg: It would be even better to have a `2.x` that has both the deprecated call and the new one. Might be possible in TypeScript with method overloads :thinking: The way I see it:

**Scenario 1 - Browser** (99% use case): He has no typings, so `Promise<string>` won't ever happen and nobody will notice a thing

**Scenario 2 - NodeJS**: Maybe TypeScript, then you get `string | null | Promise<string | null>`, and you add `await` and that's it. (confirmed by my local tests: `const a: string | null = await store.getItem()`

So the only way it would break is if you are using the React Native runtime and explicitly cast it into a "non-promise". 

I think we are good with that :smile:

src/storage.ts
francoislg: We could add some defensive code in the library to detect whether the storage is compatible in React-Native, something **very** specific like:

```
coveo.analytics.js requires an asynchronous storage solution when using it in React-Native.

You must import `@react-native-async-storage/async-storage` and initialize your client with the following code:

import AsyncStorage from '@react-native-async-storage/async-storage'; 

const analytics = new CoveoAnalytics({
  storage: new AsyncStorage()
});

Here is a link for more information: http://...
```

So we don't load the library, but it is very explicit about what must be done, and it won't be buried in the documentation :smile: 

I saw similar errors in other libraries and liked that very much

nrawji:
Added code for deprecation, I still prefer a major version bump because there is not way to support promises within the deprecated methods so we have to check if the return type from storage is a promise everywhere. I think the best solution would be to create a version `3.0.0` which is built to better support async use cases. This will also probably provide me with more flexibility to come up with a better way to handle the dependency requirements for React Native and other environments.

src/client/analytics.ts
```diff
@@ -182,7 +204,18 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         const processCustomParameters: ProcessPayloadStep = (currentPayload) =>
             usesMeasurementProtocol ? this.processCustomParameters(currentPayload) : currentPayload;
 
-        const payloadToSend = [
+        type AsyncProcessPayloadStep = (currentPayload: any) => Promise<any>;
+        const addVisitorIdStep: AsyncProcessPayloadStep = async (currentPayload) => ({
```
francoislg, line 208: The steps were in the order they appeared in the `payloadToSend` pipeline :P 

Why did you bring them down here?

francoislg (approved):

package.json
```diff
@@ -21,6 +21,9 @@
     },
     "license": "MIT",
     "devDependencies": {
+        "react": "^17.0.1",
```
samisayegh, line 24: I do not see the `react` dependency being imported in the files making up this PR. Is it needed?

package.json
```diff
@@ -21,6 +21,9 @@
...
+        "react-native": "^0.63.3",
+        "@react-native-async-storage/async-storage": "^1.13.2",
```
samisayegh, line 26: I believe `react-native` and `@react-native-async-storage/...` should be under `dependencies` since they are used at run-time? versus e.g. typescript and jest which are not.

src/client/analytics.ts
```diff
@@ -112,18 +114,17 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
         this.analyticsFetchClient = new AnalyticsFetchClient(clientsOptions);
-
-        this.initVisitorId();
     }
 
     private initRuntime(clientsOptions: IAnalyticsBeaconClientOptions) {
         if (hasWindow() && hasDocument()) {
             return new BrowserRuntime(clientsOptions, () => this.flushBufferWithBeacon());
+        } else if (isReactNative()) {
+            console.warn(ReactNativeRuntimeWarning);
```
samisayegh, line 125: Should the code return here to avoid returning the nodejs runtime?

src/client/analytics.ts
```diff
@@ -134,15 +135,45 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+                } else {
+                    this.visitorId = existingVisitorId;
+                }
+            } catch (err) {
+                console.log(
+                    'Could not get visitor ID from the current runtime environment storage. Using a random ID instead.',
+                    err
+                );
+                this.visitorId = uuidv4();
+            }
```
samisayegh, line 153: I find this `try/catch` block hard to follow, partly because the visitorId can be set in one of three places, either by the function on line 143, on line 145 or in the `catch` on line 152. I think it would be simpler to divide the logic into two steps:

1. Determine the id to use
2. Set the visitorId.

```
private async determineVisitorId() {
    try {
        return await this.storage.getItem('visitorId');
    } catch(e) {
        //log error
        return uuidv4()
    }
}

async getCurrentVisitorId() {
    if (!this.visitorId) {
        const id = await determineVisitorId()
        await this.setCurrentVisitorId(id)
    }

    return this.visitorId;
}
```

samisayegh (approved):

louis-bompart:
Hey, question, what's the impact on the size of the compiled lib?

nrawji (commented):
> Hey, question, what's the impact on the size of the compiled lib?

@louis-bompart not too much gzipped size goes from around 9.9kb on master to 10.5kb in this branch. Just to re-iterate none of the react-native specific stuff is 

> Hey, question, what's the impact on the size of the compiled lib?

It isn't a huge difference from about 9.9kb to 10.5kb.

package.json
samisayegh: I see hmm, so the exports in this file are not publicly accessible?

https://github.com/coveo/coveo.analytics.js/pull/226/files#diff-4c2d9864e94cbc3db4064814e506427432083f2b645e1ed41156f2a7ebb14d1cR1

I assumed they were intended to be imported and passed on initialization? If they are just for UTs, is an index file necessary?

If they are meant to be imported however, then I think it's a nicer experience for them not to need to install the dependencies. There should be no effect on bundle size for non-react-native users because the bundlers would tree-shake these unused exports.

src/client/analytics.ts
nrawji: I personally think its fine to have the nodejs runtime be a fallback environment.

src/client/analyticsRequestClient.ts
```diff
@@ -1,7 +1,8 @@
 import {AnyEventResponse, IRequestPayload} from '../events';
 
 export interface VisitorIdProvider {
-    currentVisitorId: string;
```
francoislg, line 4: technically a breaking change, can you put it optional and deprecated instead? :P

francoislg (approved):

src/client/analyticsRequestClient.ts
nrawji: I thought the visitor ID provider was completely internal?

nrawji: Yeah I looked further into this and I don't think a deprecated attribute is necessary here since:

1. I don't think it's possible for end users to instantiate a VisitorIdProvider themselves
2. Within `analytics.ts` the new methods `getCurrentVisitorId/setCurrentVisitorId` already wrap all functionality the old getters and setters have, so only using the new methods should suffice (basically if the old methods return id `123` the new methods will also return that ID).

francoislg: Thanks, all good!

PR #228, opened 2020-11-14
Ensure code follows prettier opinionated format, always.

louis-bompart:
Split conscientiously into multiple commits to ease the review.
_Starts opinionated word counter._

### Goals:
 - As a contributor, I don't want to deal with code style conformity.
 - As a contributor, I want to be able to code on any kind of development setup easily.

### Relevant changes:
 - Update Prettier to the latest version. Resource for review: https://prettier.io/blog/2020/03/21/2.0.0.html (there's some gotcha)
 - Update tsjs.
 - Create a `.prettierrc.js` for better extendability.
 - Strip down .prettierignore: now it's a copy 1-1 of the .gitignore. This is an important IMO because it ensures that everything that can be committed is opiniatedly check.
 - Add lint-staged: a neat tool that'll prettify your code for you before sending it to the remote, so you just don't have to bother thinking about it.
 - Add CI-check: rules are nice, but they need to be enforced. This does that.
 - Replaced UNIX `rm` by `rimraf`: Package is already in the dependency tree, so it's really a 0 cost change.

### Irrelevant changes:
https://github.com/coveo/coveo.analytics.js/commit/82e92e853adb58f81e2916aa394f9c486301c37d is the result of running prettier and committing, without any other kind of frill :soopinionated:.

### Possible future improvements:
 - If some rules bother or others should be introduced, well it can be done. I purposefully didn't do any of that.
 - ESLint could be nice.

### References:
- https://prettier.io/docs/en/install.html#summary

---
https://coveord.atlassian.net/browse/SFINT-3567


louis-bompart:
See https://github.com/coveo/coveo.analytics.js/pull/229 for what happens in the CI when someone does not follow the ruleset.

public/index.html
```diff
@@ -1,34 +1,55 @@
 <html>
-<script>
-(function(c,o,v,e,O,u,a){
-a='coveoua';c[a]=c[a]||function(){(c[a].q=c[a].q|| []).push(arguments)};
-c[a].t=Date.now();u=o.createElement(v);u.async=1;u.src=e;
-O=o.getElementsByTagName(v)[0];O.parentNode.insertBefore(u,O)
-})(window,document,'script','coveoua.js')
+    <script>
```
francoislg, line 2: This file here should definitely be ignored.

This specific script is intended to be copy-pasted as-is, and I think it obfuscates it a bit compared to the documentation

francoislg (approved):
Looks good

cnguyen-coveo (approved):

public/index.html
louis-bompart: I adressed your concern by prettier-ignore the snippet, not the whole file, just wanted to be sure it's 👍 with you so I re-requested a review ;)

.mailmap
```diff
@@ -1,9 +1,9 @@
...
-Pierre-Alexandre St-Jean <pa@stjean.me> <pastjean@coveo.com>
+# Generate CONTRIBUTORS.md: contributors.sh
+
+# Tip for finding duplicates (besides scanning the output of CONTRIBUTORS.md for name
+# duplicates that aren't also email duplicates): scan the output of:
+#   git log --format='%aE - %aN' | sort -uf
+#
+# For explanation on this file format: man git-shortlog
+
+Pierre-Alexandre St-Jean <pa@stjean.me> <pastjean@coveo.com>
```
dlafreniere, line 9: I think we can ditch this file :rip: pastjean

public/index.html
francoislg: all good

dlafreniere (approved):

PR #229, opened 2020-11-14
#228, but not actually prettified

louis-bompart:
Demo for #228.
Demonstrate what happens if a contributor does not lint its code (here, I just reverted https://github.com/coveo/coveo.analytics.js/commit/82e92e853adb58f81e2916aa394f9c486301c37d, but anyone can do anything by committing with the `--no-verify` flag.)

Of course, do not merge this one.


louis-bompart:
#228 merged ✔️

PR #230, opened 2020-11-20
Add beforeSendHooks to AnalyticsClient option

olamothe:
For Headless, we want to expose the possibility for end users to augment analytics event with metadata that is unique to their setup.

For example, they might want to add a metadata that `userDepartment = admin/marketing/sales/r&d` to all events sent for their search page.

By exposing the before send hooks as an option on AnalyticsClient (and some plumbing done in headless), they should be able to modify the payload to add metadata if needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.12.0",
+    "version": "2.13.0",
     "description": "📈 Coveo analytics client (node and browser compatible) ",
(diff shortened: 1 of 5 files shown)


francoislg (approved):
Simple enough! :)

ThibodeauJF (approved):

nrawji (approved):

samisayegh (approved):

PR #231, opened 2020-11-23
feat(SFINT-3553): Add basic Service UA Plugin

louis-bompart:
Goal: Add support for Service Ticket Data and Service Events (without breaking anything:tm:)

Test:
 - UT
 - FT
 - E2E SAN-Check (run service.html and check the data reach the data lake)

The idea of this PR is to set a first simple stone in a 'multi-plugin' environment and see what could be done from there.
However, the idea of supporting multiple plugins is too tall of an order to be part of the scope of SFINT-3553.

Therefore here are some next steps/improvements:
 - For CA.js project, refactor the code in common between ec and svc. The goal of this refactor would be to isolate the product/commerce data and processing in ec, the ticket/service data and processing in svc and isolate the 'commons' in new plugins
 - Refactor the 'common' method of a plugin to make a plugin helper that could be used by ec, svc, and future plugins.


rollup.config.js
```diff
@@ -26,6 +26,9 @@ const browser = {
                   contentBase: ['dist', 'public'],
                   port: 9001,
                   open: true,
+                  headers: {
+                      'Access-Control-Allow-Origin': 'http://localhost:9001',
+                  },
```
louis-bompart, line 31: Fun stuff here, this is required to avoid having CORS and other fun browser security features blowing up on your face. I'd argue that it'd be nice to keep it.

src/client/measurementProtocolMapper.ts
```diff
@@ -140,25 +173,32 @@ const convertImpressionToMeasurementProtocol = (
...
+const ticketKeyRegex = new RegExp(`^(${ticketSubKeysMatchGroup}$)`);
 const productKeyRegex = new RegExp(`^${productPrefixMatchGroup}(${productSubKeysMatchGroup})$`);
 const impressionKeyRegex = new RegExp(`^(${impressionPrefixMatchGroup}(${impressionSubKeysMatchGroup}))|(il[0-9]+nm)$`);
 const customProductKeyRegex = new RegExp(`^${productPrefixMatchGroup}custom$`);
 const customImpressionKeyRegex = new RegExp(`^${impressionPrefixMatchGroup}custom$`);
 
+const isTicketKey = (key: string) => ticketKeyRegex.test(key);
 const isProductKey = (key: string) => productKeyRegex.test(key);
 const isImpressionKey = (key: string) => impressionKeyRegex.test(key);
+const is = (key: string) => productKeyRegex.test(key);
```
dlafreniere, line 194: oops

src/coveoua/simpleanalytics.ts
```diff
@@ -32,6 +33,7 @@ export class CoveoUA {
 
         if (this.client) {
             this.plugins.register('ec', new EC({client: this.client}));
+            this.plugins.register('svc', new SVC({client: this.client}));
```
dlafreniere, line 36: we originally registered the EC plugin because it was the only one. Now I'm not sure if we want to automatically register everything by default or have the client explicitly choose.
@francoislg, thoughts?

src/client/measurementProtocolMapper.ts
```diff
@@ -140,25 +173,32 @@ const convertImpressionToMeasurementProtocol = (
...
 );
+const ticketKeysMappingValues = keysOf(ticketKeysMapping).map((key) => ticketKeysMapping[key]);
 const productKeysMappingValues = keysOf(productKeysMapping).map((key) => productKeysMapping[key]);
 const impressionKeysMappingValues = keysOf(impressionKeysMapping).map((key) => impressionKeysMapping[key]);
 
+const ticketSubKeysMatchGroup = [...ticketKeysMappingValues].join('|');
 const productSubKeysMatchGroup = [...productKeysMappingValues, 'custom'].join('|');
 const impressionSubKeysMatchGroup = [...impressionKeysMappingValues, 'custom'].join('|');
 const productPrefixMatchGroup = '(pr[0-9]+)';
 const impressionPrefixMatchGroup = '(il[0-9]+pi[0-9]+)';
```
dlafreniere, line 184: you may need to change this one as well, the `pi` prefix is hardcoded

src/plugins/svc.ts
```diff
@@ -0,0 +1,246 @@
...
+const allSVCEventTypes = Object.keys(SVCPluginEventTypes).map(
+    (key) => SVCPluginEventTypes[key as keyof typeof SVCPluginEventTypes]
+);
+
+export type CustomValues = {
+    [key: string]: string | number | boolean;
+};
+
+export interface TicketProperties {
+    id?: string;
```
dlafreniere, line 24: not mandatory?

src/plugins/svc.ts
```diff
@@ -0,0 +1,246 @@
...
+        this.impressions = [];
+        this.action = undefined;
+        this.actionData = {};
+    }
+
+    clearTicket() {
+        this.ticket = {};
+    }
+
+    private addHooksForSVEvents() {
```
dlafreniere, line 97: SVC?

src/plugins/svc.ts
```diff
@@ -0,0 +1,246 @@
...
+    private addHooksForEvent() {
+        this.client.addEventTypeMapping(SVCPluginEventTypes.event, {
+            newEventType: EventType.collect,
+            variableLengthArgumentsNames: ['eventCategory', 'eventAction', 'eventLabel', 'eventValue'],
+            addVisitorIdParameter: true,
+            usesMeasurementProtocol: true,
+        });
+    }
+
+    private addSVDataToPayload(eventType: string, payload: any) {
```
dlafreniere, line 121: SVC?

src/plugins/svc.ts
```diff
@@ -0,0 +1,246 @@
...
+                };
+            }, {});
+    }
+
+    private assureBaseImpressionValidity(baseImpression: BaseImpression) {
+        const {position, ...baseImpressionRest} = baseImpression;
+        if (position !== undefined && position < 1) {
+            console.warn(
+                `The position for impression '${baseImpression.name || baseImpression.id}'` +
+                    ` must be greater than 0 when provided.`
```
dlafreniere, line 170: I don't know if the formatter split this string, but _technically_ the second half doesn't need backticks

louis-bompart: You're right, will correct that 👍 
Prettier _never_ split a string, regardless of the kind of used quotes afaik.
So it should be
```ts
console.warn(
    `The position for impression '${baseImpression.name || baseImpression.id}' must be greater than 0 when provided.`
)```

File: src/plugins/svc.ts (hunk 5 above)
Comment (louis-bompart) [reply src/plugins/svc.ts:24]: Yes, this is 'as designed':
When a Ticket is sent to the UA, the client has not necessarily created a ticket, it can still be in a form, filling up stuff, or being offered suggestions. Therefore, we cannot always provide an ID for it.

We toyed with the idea of having a `ticket_provisional_id`, which would be generated by the CA.js through a custom function of the plugin.
This ID would have been unique during the whole time the end-user was in the context of this 'creation-in-progress' ticket, more precisely:
 - It would have been created as soon as the user enters an 'SVC-powered ticket process' (i.e. some kind of user flow where the user deals with a ticket or part of it, and where the ServiceAPI does intervene).
 - It would have been 'ditched' as soon as the user either abandon the flow or complete it. To complete it, a specific event would have been sent, with the `id` l24 and this `ticket_provisional_id`.
This would have allowed to technically be able to 'stitch' the id together (i.e. ok, when I'm dealing with `ticket_provisional_id: 12345`, it's the same as `ticket_id: whatever4586`

In the end, we chose to not go forward with this solution, because we can already achieve our goal of being able to 'track' what the user was doing before creating an actual ticket using the `pageviewid`, `visitorId` etc...


So the rule of thumb with TicketProperties in the end 'send what you have'

File: src/coveoua/simpleanalytics.ts (hunk 3 above)
Comment (louis-bompart) [reply src/coveoua/simpleanalytics.ts:36]: My suggestion would be a combination of both?
Having a default set of plugins, but allow the user to specify its own set.

The question is then, should SVC be part of the default set.
(I think EC should be, or it'll be a breaking change)

File: src/client/measurementProtocolMapper.ts (hunk 4 above)
Comment (louis-bompart) [reply src/client/measurementProtocolMapper.ts:184]: 👍 Probably, but not on this one, we decided for our initial phase to log the 'impression' server-side, to get a toe in the water.

By server-side impression, I mean that for now what we decided to do is, when a request is received, just before answering it, the backend will log a UA with the request payload, the visitorId (it's a required param for the service endpoints) and the response payload.
We then assume (at least for now) that all's that's returned by the service is 'printed' (so impression ✔️ ), no questions asked.

TL;DR: For now, we don't need a client-side impression, so I would not touch this 😄

File: src/plugins/svc.ts [hunk 9]
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    private addHooksForSVEvents() {
+        this.client.registerBeforeSendEventHook((eventType, ...[payload]) => {
+            return allSVCEventTypes.indexOf(eventType) !== -1 ? this.addSVDataToPayload(eventType, payload) : payload;
+        });
+    }
+
+    private addHooksForPageView() {
+        this.client.addEventTypeMapping(SVCPluginEventTypes.pageview, {
```
francoislg, line 104: If you check here: https://github.com/coveo/coveo.analytics.js/blob/master/src/client/analytics.ts#L268

you can see that the map is actually unique by event type (the first parameter). So here, if you include the `SVC` plugin after `EC`, it will override the EC configuration :confused: 

Not sure how to easily fix that, though... We could just delete `addEventTypeMapping` and hardcode the configuration in the client so that it is common to every plugin

francoislg (commented):
Looking forward for this phase 2 refactor, looking at the replicated `plugins/svc.ts` makes me a bit sad :wink:

src/plugins/svc.ts
louis-bompart: Indeed, I'd argue that, because the hooks configurations are identical between SVC and EC, it should be part of the 'refactor later'.

src/client/measurementProtocolMapper.ts
```diff
@@ -17,6 +18,20 @@ const productKeysMapping: {[key in keyof Product]: string} = {
     position: 'ps',
 };
 
+const ticketKeysMapping: {[key in keyof Ticket]: string} = {
```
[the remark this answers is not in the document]
louis-bompart: Yeah this doesn't go against the main reason why I've put refactor in a future phase, so will do 👍

louis-bompart:
> Looking forward for this phase 2 refactor, looking at the replicated `plugins/svc.ts` makes me a bit sad 😉

No worries @francoislg ain't liking much either 😓 .
But my goal here was to avoid touching EC as much as possible to 'ease' the concern of the new plugin entering the ring.

Also, I think it'd be best to sit and think properly on how to 'plugify' CA.js on its own story and without having a specific implementation/usecase in mind.

dlafreniere (approved):

louis-bompart:
A question that I had that I did not answer today was, can I move the click to the measurementProtocol, and what concern should I have about it?

i.e.

What the concerns of doing this:
```ts
        this.client.addEventTypeMapping(SVCPluginEventTypes.click, {
            newEventType: EventType.collect,
            variableLengthArgumentsNames: [/*tbd*/],
            addVisitorIdParameter: true,
            usesMeasurementProtocol: true,
        });
```

francoislg:
> A question that I had and did not answer today was with this PR, can I move the click to the measurementProtocol, and what concern should I have about it?

There are no "clicks" in the measurement protocol. The pageview should be sufficient

I wonder what is your use case for that, but I think it would be best to discuss this in Slack or something :P

francoislg (approved):

PR #233, opened 2020-11-30
Allow non predefined custom event type

olamothe:
The current `logCustomEvent` function tries to find a custom event type from the predefined map in `searchPageEvents.ts`.

This is helpful for all events we want to log out of the box. 

But, if someone wanna create a new type of custom event, it needs to be possible to do so by creating your own event value and event type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.13.0",
+    "version": "2.14.0",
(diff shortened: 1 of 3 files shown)


dlafreniere (commented):
Tests?

ThibodeauJF (approved):

samisayegh (approved):

PR #234, opened 2020-12-01
[COM-623] Lowercase custom properties

agagnon-coveo:
Ensure that the custom objects passed from plugins (such as ec and svc) are consistent with the other properties from the measurement protocol which are all lowercased.

Still need to test side effects and see if typing in the `ensureLowercaseKeys` method could be improved.

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -59,7 +59,7 @@ describe('ec events', () => {
         });
     });
 
-    it('should fix the cashing on custom objects', async () => {
+    it('should lowercase the properties of custom objects', async () => {
         await coveoua('set', 'custom', {
             camelCase: 'camel',
             snake_case: 'snake',
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -59,6 +59,27 @@ describe('ec events', () => {
         });
     });
 
+    it('should fix the cashing on custom objects', async () => {
+        await coveoua('set', 'custom', {
+            camelCase: 'camel',
+            snake_case: 'snake',
+            PascalCase: 'pascal',
+            UPPER_CASE_SNAKE: 'upper snake',
+        });
+        await coveoua('send', 'pageview');
+
+        const [body] = getParsedBody();
+
+        expect(body).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            camelcase: 'camel',
+            snake_case: 'snake',
+            pascalcase: 'pascal',
+            upper_case_snake: 'upper snake',
+        });
+    });
+
     it('can send a product detail view event', async () => {
         coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', unknown: 'ok'});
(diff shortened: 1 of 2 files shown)


src/client/analytics.ts
```diff
@@ -330,14 +330,28 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
             ...newPayload,
         };
     }
 
+    private ensureLowercaseKeys(custom: any) {
+        const keys = Object.keys(custom || {});
+
+        const lowercasedCustom = {} as {[key: string]: any};
+
+        keys.forEach((key) => {
```
francoislg, line 348: I think this could be changed to a `reduce`: 

```
return keys.reduce((all, key) => ({ 
  ...all, 
  [key.toLowerCase()]: custom[key] 
}), {});
```

francoislg (approved):
Looks like it :)

Spuffynism (approved):

functional/ec-events.spec.ts
```diff
@@ -59,6 +59,27 @@ describe('ec events', () => {
         });
     });
 
+    it('should fix the cashing on custom objects', async () => {
```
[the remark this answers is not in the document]
toussaints: I think I would even mention the fact that it lowercase in the title of the test
>should lowercase keys on custom objects

src/client/analytics.ts
```diff
@@ -330,14 +330,28 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
         const newPayload = convertCustomMeasurementProtocolKeys(rest);
 
         return {
-            ...(custom || {}),
+            ...lowercasedCustom,
             ...newPayload,
         };
     }
 
+    private ensureLowercaseKeys(custom: any) {
```
toussaints, line 343: not sure about the `ensure` word
`lowercaseKeys` would have done it I think

functional/ec-events.spec.ts
```diff
@@ -59,6 +59,27 @@ describe('ec events', () => {
...
+        await coveoua('set', 'custom', {
+            camelCase: 'camel',
+            snake_case: 'snake',
+            PascalCase: 'pascal',
+            UPPER_CASE_SNAKE: 'upper snake',
```
dlafreniere, line 67: no love for `kebab-case`? 🍢

toussaints (approved):

src/client/analytics.ts
agagnon-coveo: Yeah, good idea, it is superfluous and not totally accurate anyways.

functional/ec-events.spec.ts
agagnon-coveo: It's only for casing and there's already a case to make sure special characters are handled properly (with snake_case) :grin:

functional/ec-events.spec.ts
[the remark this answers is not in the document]
agagnon-coveo: Woops :money_with_wings:

francoislg (approved):

PR #235, opened 2020-12-08
Add support for facet-state in search event request

olamothe:
For DynamicFacet experience, UA (and ML) need to have access to the full "facet state" when a facet value is selected/deselected.

This adds the new definition expected by UA 

https://platformdev.cloud.coveo.com/docs?urls.primaryName=Usage%20Analytics%20Write#/Analytics%20API%20-%20Version%2015/post__v15_analytics_search

Also modifies the different "logFacet" function in SearchPageClient to request that new information.

### DIFF
--- a/src/events.ts
+++ b/src/events.ts
@@ -37,6 +37,18 @@ export interface EventBaseRequest {
     originLevel3?: string;
 }
 
+export interface FacetStateRequest {
+    field: string;
+    id: string;
+    value: string;
+    valuePosition: number;
+    displayValue: string;
+    facetType: 'specific' | 'dateRange' | 'numericalRange' | 'hierarchical';
+    state: 'idle' | 'selected';
+    facetPosition: number;
+    title: string;
+}
+
 export interface SearchEventRequest extends EventBaseRequest {
     searchQueryUid: string;
     queryText: string;
@@ -48,6 +60,7 @@ export interface SearchEventRequest extends EventBaseRequest {
     results?: SearchDocument[];
     queryPipeline?: string;
     userGroups?: string[];
+    facetState?: FacetStateRequest[];
 }
 
 export interface DocumentInformation {
--- a/src/events.ts
+++ b/src/events.ts
@@ -44,9 +44,12 @@ export interface FacetStateRequest {
     valuePosition: number;
     displayValue: string;
     facetType: 'specific' | 'dateRange' | 'numericalRange' | 'hierarchical';
-    state: 'selected';
+    state: 'selected' | 'idle';
     facetPosition: number;
     title: string;
+    start?: string;
+    end?: string;
+    endInclusive?: boolean;
 }
 
 export interface SearchEventRequest extends EventBaseRequest {
--- a/src/events.ts
+++ b/src/events.ts
@@ -44,7 +44,7 @@ export interface FacetStateRequest {
     valuePosition: number;
     displayValue: string;
     facetType: 'specific' | 'dateRange' | 'numericalRange' | 'hierarchical';
-    state: 'idle' | 'selected';
+    state: 'selected';
     facetPosition: number;
     title: string;
 }
--- a/src/searchPage/searchPageClient.spec.ts
+++ b/src/searchPage/searchPageClient.spec.ts
@@ -1,6 +1,12 @@
 import * as fetchMock from 'fetch-mock';
 import {CoveoSearchPageClient} from './searchPageClient';
-import {SearchPageEvents, PartialDocumentInformation, DocumentIdentifier, CustomEventsTypes} from './searchPageEvents';
+import {
+    SearchPageEvents,
+    PartialDocumentInformation,
+    DocumentIdentifier,
+    CustomEventsTypes,
+    FacetStateMetadata,
+} from './searchPageEvents';
 import CoveoAnalyticsClient from '../client/analytics';
 import {NoopAnalytics} from '../client/noopAnalytics';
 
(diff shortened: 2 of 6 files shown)


src/events.ts
```diff
@@ -37,6 +37,18 @@ export interface EventBaseRequest {
...
 
+export interface FacetStateRequest {
+    field: string;
+    id: string;
+    value: string;
+    valuePosition: number;
+    displayValue: string;
+    facetType: 'specific' | 'dateRange' | 'numericalRange' | 'hierarchical';
+    state: 'idle' | 'selected';
+    facetPosition: number;
```
samisayegh, line 48: Hmm, not related to this PR, but I am thinking about how headless would determine this. We could use the position of the facet in the response array, but if the interface is not using the facet manager, then this may not reflect what the user is seeing/layout of the page.

src/searchPage/searchPageClient.spec.ts
```diff
@@ -23,6 +29,20 @@ describe('SearchPageClient', () => {
...
     };
 
+    const fakeFacetState = [
+        {
+            valuePosition: 0,
+            value: 'foo',
+            state: 'selected' as const,
+            facetPosition: 1,
+            displayValue: 'foobar',
+            facetType: 'specific' as const,
```
samisayegh, line 39: Why is `as const` needed here? Does the value need to be immutable?

src/searchPage/searchPageClient.ts
```diff
@@ -100,8 +101,11 @@ export class CoveoSearchPageClient {
         return this.logSearchEvent(SearchPageEvents.searchboxAsYouType);
     }
 
-    public logBreadcrumbFacet(metadata: FacetMetadata | FacetRangeMetadata | CategoryFacetMetadata) {
-        return this.logSearchEvent(SearchPageEvents.breadcrumbFacet, metadata);
+    public logBreadcrumbFacet(
```
samisayegh, line 104: Could a better name for this method be `logBreadcrumbSearch` or `logBreadcrumbClick`? Having "facet" in the name feels strange,

samisayegh (approved):

ThibodeauJF (approved):

src/events.ts
ThibodeauJF: True, it won't be simple, no way to know implicitly from the headless perspective

src/searchPage/searchPageClient.spec.ts
olamothe: Yes.
```
type something = 'foo' | 'bar';
const myVar: something = 'foo' // ok for TypeScript, myVar cannot change, so it knows the type is correct.
var myVar: something = 'foo' // not ok for TypeScript, myVar can be changed to any string.
```

src/events.ts
olamothe: For now, I'm only going to look at the position in response array. 

But you are right, this won't "always" work, and we'll most probably have to revisit at a later time 🤷

src/searchPage/searchPageClient.ts
```diff
@@ -287,6 +280,32 @@ export class CoveoSearchPageClient {
...
+    ) {
+        const payload = {
+            ...this.getBaseSearchEventRequest(event, metadata),
+            facetState,
+        };
+
+        return this.coveoAnalyticsClient.sendSearchEvent(payload);
+    }
+
+    private getBaseSearchEventRequest(event: SearchPageEvents, metadata?: Record<string, any>) {
```
francoislg, line 296: I think it would make sense to type the return as `SearchEventRequest`, just to ensure that this method returns all the required properties and not just an anonymous type that happens to match the request.

src/searchPage/searchPageClient.ts
```diff
@@ -100,8 +101,11 @@ export class CoveoSearchPageClient {
         return this.logSearchEvent(SearchPageEvents.searchboxAsYouType);
     }
 
-    public logBreadcrumbFacet(metadata: FacetMetadata | FacetRangeMetadata | CategoryFacetMetadata) {
-        return this.logSearchEvent(SearchPageEvents.breadcrumbFacet, metadata);
+    public logBreadcrumbFacet(
+        metadata: FacetMetadata | FacetRangeMetadata | CategoryFacetMetadata,
+        facetState: FacetStateMetadata[]
```
francoislg, line 106: Technically a breaking change :thinking: 

Should all `facetState` parameters here be put optional? And fallback on the usual search event?

src/searchPage/searchPageClient.ts
```diff
@@ -162,36 +166,36 @@ export class CoveoSearchPageClient {
...
+        return this.logFacetSearchEvent(SearchPageEvents.facetSelectAll, meta, facetState);
     }
 
-    public logFacetUpdateSort(meta: FacetSortMeta) {
-        return this.logSearchEvent(SearchPageEvents.facetUpdateSort, meta);
+    public logFacetUpdateSort(meta: FacetSortMeta, facetState: FacetStateMetadata[]) {
+        return this.logFacetSearchEvent(SearchPageEvents.facetUpdateSort, meta, facetState);
     }
 
     public logFacetShowMore(meta: FacetBaseMeta) {
```
francoislg, line 201: Just to be sure, those two `logFacet` calls don't need the new metadata?

src/searchPage/searchPageClient.ts
olamothe: Indeed, technically a breaking change. 
But, headless is the only project using this class here ;) 

So, yeah, in theory we should not do this, but in practice, it only impact us. 
I'd prefer not to make it optional.

src/searchPage/searchPageClient.ts
olamothe: No, those are custom events, and we don't need it in those cases.

olamothe:
I will be pushing more changes to the `FacetStateRequest` interface, I realized it's missing some properties needed for range facets (start/end).

src/events.ts
olamothe: I've added a comment in PR: https://github.com/coveo/ui-kit/pull/346 regarding this.

src/searchPage/searchPageClient.ts
olamothe: Will do !

PR #237, opened 2020-12-29
chore(SFINT-3596): Refactor Plugins

louis-bompart:
I refactored the plugins and measurement protocol mapping.

 I created a 'base' plugin that is in charge of logging the minimal data required by the Measurement Protocol.
 The other plugins extend upon it, this ensures that every plugin does log the minimal data required (and also simplify the job of a plugin implementer, because he just has to focus on the data relevant to his plugin)
 
 The measurement protocol is divided similarly, with three mapping files: one for the 'base', one for service, one for commerce, and finally, a mapper that consume the three mapping files into single functions, later used by analytics.ts
 
 And ride-along come the deletion of the impression in the service plugin. We did not implement client-side impressions yet and the type definition and the method could lead to confusion.

### DIFF
--- a/src/client/measurementProtocolMapper.ts
+++ b/src/client/measurementProtocolMapper.ts
@@ -51,13 +51,13 @@ export const convertCustomMeasurementProtocolKeys = (data: {[name: string]: stri
     }, {});
 };
 
-const getFirstCustomMeasurementProtocolKeyMatch = (key: string): string | null => {
-    let match = null;
+const getFirstCustomMeasurementProtocolKeyMatch = (key: string): string | undefined => {
+    let matchedKey: string | undefined = undefined;
     [...isCustomCommerceKey].every((regex) => {
-        match = regex.exec(key);
-        return !Boolean(match);
+        matchedKey = regex.exec(key)?.[1];
+        return !Boolean(matchedKey);
     });
-    return match;
+    return matchedKey;
 };
 
 const convertCustomObject = (prefix: string, customData: {[name: string]: string}) => {
--- a/src/client/measurementProtocolMapper.ts
+++ b/src/client/measurementProtocolMapper.ts
@@ -1,5 +1,9 @@
 import {isServiceKey, serviceActionsKeysMapping} from './measurementProtocolMapping/serviceMeasurementProtocolMapper';
-import {commerceActionKeysMapping, isCommerceKey, isCustomCommerceKey} from './measurementProtocolMapping/commerceMeasurementProtocolMapper';
+import {
+    commerceActionKeysMapping,
+    isCommerceKey,
+    isCustomCommerceKey,
+} from './measurementProtocolMapping/commerceMeasurementProtocolMapper';
 import {keysOf} from './utils';
 import {baseMeasurementProtocolKeysMapping} from './measurementProtocolMapping/baseMeasurementProtocolMapper';
 
--- a/src/client/measurementProtocolMapper.ts
+++ b/src/client/measurementProtocolMapper.ts
@@ -1,87 +1,12 @@
-import {EC, Product, ImpressionList, BaseImpression} from '../plugins/ec';
-import {isTicketKey, svcActionsKeysMapping} from './coveoServiceMeasurementProtocolMapper';
+import {isServiceKey, serviceActionsKeysMapping} from './measurementProtocolMapping/serviceMeasurementProtocolMapper';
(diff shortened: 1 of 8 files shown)


src/client/measurementProtocolMapper.ts
```diff
@@ -185,6 +51,16 @@ export const convertCustomMeasurementProtocolKeys = (data: {[name: string]: stri
     }, {});
 };
 
+const getFirstCustomMeasurementProtocolKeyMatch = (key: string): string | undefined => {
+    const customKeyRegExps: RegExp[] = [...isCustomCommerceKey];
+    for (let index = 0; index < customKeyRegExps.length; index++) {
```
francoislg, line 56: I think this loop could be done with a simple find?

```
const firstMatch = customKeyRegExps.find(regex =>  customKeyRegExps[index].exec(key));
return firstMatch?.[1];
```

I find it slightly more expressive, but feel free to keep the loop (or adapt my version to your liking)

src/plugins/ec.ts
```diff
@@ -92,16 +81,10 @@ export class EC {
         this.impressions.push(impression);
     }
 
-    setAction(action: string, options?: any) {
-        this.action = action;
-        this.actionData = options;
-    }
-
     clearData() {
```
francoislg, line 84: Should this instead be an abstract method as well?

Like, in the base plugin:

```
protected abstract clearPluginData();

public clearData() {
  this.action = undefined;
  this.actionData = {};
  this.clearPluginData();
}
```

This way, we don't have to bother about the `super.clearData` and it makes it very explicit when implementing a new plugin that you need to clear some data :smile:

src/plugins/svc.spec.ts
```diff
@@ -137,7 +137,6 @@ describe('SVC plugin', () => {
 
     it('should be able to clear all the data', () => {
         svc.setTicket({subject: '🍨', description: 'Some desc'});
-        svc.addImpression({id: '🍦', name: 'impression'});
```
francoislg, line 140: :confused: I don't understand why this was removed

francoislg (approved):
Looks very good! Nice job!

src/plugins/svc.spec.ts
louis-bompart: We didn't spend anytime 'Researching and Designing' client-side impression for service content yet, and it was a mistake from my part to actually include impression in the CA.js plugin for service at this stage because of that.
We will eventually add it back, but because currently we're unsure about how we want to do it, I'd rather not have it possible with the lib.

What do you think?

src/plugins/ec.ts
louis-bompart: > Should this instead be an abstract method as well?
> 
> Like, in the base plugin:
> 
> ```
> protected abstract clearPluginData();
> 
> public clearData() {
>   this.action = undefined;
>   this.actionData = {};
>   this.clearPluginData();
> }
> ```
> 
> This way, we don't have to bother about the `super.clearData` and it makes it very explicit when implementing a new plugin that you need to clear some data 😄

I thought about it, but in the end I think that it make sense that `action` and `actionData` to be a concern of the Base Plugin class and to 'enforce' the cleaning as a mandatory behavior for a plugin (so that plugin behaviors are consistent, to avoid confusion for the user of the lib).

What do you think?

src/client/measurementProtocolMapper.ts
louis-bompart: Thanks, that's way better 👍

src/plugins/ec.ts
francoislg: Yeah, see my design, `action` and `actionData` are still a concern of the Base Plugin :smile: 

My suggestion is just from a programming standpoint: 

Forcing to implement a `clearPluginData` method (my code)
VS 
Asking to *extend* an existing `clearData` method (the current code)

I think the first behavior is more explicit

src/plugins/svc.spec.ts
dlafreniere: makes sense

dlafreniere (approved):

src/plugins/ec.ts
louis-bompart: Ohhhh I misread. You're right!

src/client/measurementProtocolMapper.ts
francoislg: "kinda inefficient" does not really matter at this scale, to be honest ;) 

if it was a thousand regex or so, maybe, but let's say we have 10 plugins, it still won't be a problem

Your call, but the code clarity outweighs the performance gain to me :wink:

PR #245, opened 2021-01-07
feat(SFINT-3665): Add ability to select plugin on CA.js initialization.

louis-bompart:
- Move plugin management into `plugin.ts`.
- (Very) strongly type the plugins stuff.
- Ensure retro-compatibility on init to avoid breaking changes.
- Reworked simpleanalytics tests:
  - Added some more precise assertion on tests. This revealed that some tests were not testing what they say they were testing.
  - Fixed the tests that were using the Express Server: they reported as 'working' falsely because they were not awaited on. It was less of a pain to ditch the Express server altogether and uses fetch-mock, so I did so (and IMO it's more readable like so)
  - Added more tests for some untested case, was not costing much more and that allows plugin.ts and simpleanalytics to reach 💯 coverage.


francoislg (changes requested):
The most important questions are related to the external plugins.

The rest is really nice, thanks for the tests refactor with fetchMock, way easier to understand!

src/coveoua/simpleanalytics.spec.ts
```diff
@@ -1,8 +1,24 @@
...
-import {Version} from '../client/analytics';
 import {createAnalyticsClientMock} from '../../tests/analyticsClientMock';
+import {EC} from '../plugins/ec';
+import {SVC} from '../plugins/svc';
+import 'isomorphic-fetch';
+import * as fetchMock from 'fetch-mock';
+
+jest.mock('../plugins/svc', () => {
+    const SVC = jest.fn().mockImplementation(() => {});
+    (SVC as any)['Id'] = 'svc';
```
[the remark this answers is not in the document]
louis-bompart: No because Id is a static property 😅

src/plugins/BasePlugin.ts
```diff
@@ -2,11 +2,18 @@ import {AnalyticsClient} from '../client/analytics';
...
+    readonly Id: string;
+};
+export type CoveoUAPlugin<T = typeof BasePlugin> = T & PluginWithId;
+
 export const BasePluginEventTypes = {
     pageview: 'pageview',
     event: 'event',
 };
 
+export type PluginOption = {client: AnalyticsClient; uuidGenerator?: typeof uuidv4};
```
dlafreniere, line 15: `PluginOptions`

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,33 @@
...
+};
+type GenericPlugin = {[fnName: string]: (...any: UAPluginOptions) => any} & BasePlugin;
 
 export class Plugins {
-    private plugins: {[name: string]: any} = {};
+    public static readonly DefaultPlugins: AvailablePluginsNames[] = [EC.Id, SVC.Id];
+
+    private registeredPlugin: Partial<Record<AvailablePluginsNames, BasePlugin>> = {};
 
-    register(name: string, plugin: any): void {
```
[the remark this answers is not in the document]
louis-bompart: I'll probably take a leaf out of Google's book again and do something similar

I'll make sure we're mo' flexible and keep on allowing external plugins. This mean also that AvailablePluginNames interface will be axed because it won't have much sense after.

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,33 @@
+import {BasePlugin, PluginOption} from '../plugins/BasePlugin';
+import {EC} from '../plugins/ec';
+import {SVC} from '../plugins/svc';
+
 export type UAPluginOptions = any[];
+export type AvailablePluginsNames = keyof typeof pluginMap;
```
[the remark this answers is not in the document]
louis-bompart: Will be removed because of the conclusion of this discussion https://github.com/coveo/coveo.analytics.js/pull/245/files/632dc3f98c9ebdd950cabc0fbb0735ea2774434f#r553934766

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,50 @@
...
 export class Plugins {
-    private plugins: {[name: string]: any} = {};
+    public static readonly DefaultPlugins: string[] = [EC.Id, SVC.Id];
+    private registeredPluginsMap: Map<string, Plugin> = (() => {
+        const map = new Map();
+        map.set(EC.Id, EC);
+        map.set(SVC.Id, SVC);
+        return map;
+    })();
+    private requiredPlugins: Partial<Record<string, Plugin>> = {};
```
francoislg, line 19: why is it Partial?

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,50 @@
...
 export type UAPluginOptions = any[];
 
+export type Plugin = {[fnName: string]: (...any: UAPluginOptions) => any} & BasePlugin & {
+        new (options: PluginOptions): Plugin;
+    };
+
 export class Plugins {
-    private plugins: {[name: string]: any} = {};
+    public static readonly DefaultPlugins: string[] = [EC.Id, SVC.Id];
+    private registeredPluginsMap: Map<string, Plugin> = (() => {
```
francoislg, line 13: Would this work instead? 
```
private registeredPluginsMap = {
 [EC.Id]: EC,
 [SVC.Id: SVC
}
```

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,50 @@
...
+        if (!pluginClass) {
+            throw new Error(
+                `No plugin named "${name}" is currently registered. If you use a custom plugin, use 'registerPlugin' first.`
+            );
+        }
+        this.requiredPlugins[name] = new pluginClass(option);
+    }
+
+    register(name: string, plugin: Plugin) {
+        this.registeredPluginsMap.set(name, plugin);
```
francoislg, line 32: :thinking: now that we have `require` mimicking `analytics.js`'s format, `register` is kind of confusing.

Maybe we could rename `register` to `provide` just like `analytics.js`? It was all internal anyway, so we might even be able to safely remove `register`

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,50 @@
...
+    }
 
-    register(name: string, plugin: any): void {
-        this.plugins[name] = plugin;
+    unrequire(): void {
```
francoislg, line 35: `clearRequired` ?

At first glance, I would think `unrequire` to be the opposite of `require` and take a `name` as a parameter :thinking:

src/coveoua/simpleanalytics.ts
```diff
@@ -89,6 +115,17 @@ export class CoveoUA {
         callback();
     }
 
+    registerPlugin(name: string, plugin: Plugin) {
```
francoislg, line 118: Should it be named `provide` to follow `analytics.js` ?

src/coveoua/plugins.ts
louis-bompart: Yeah that was the idea at first. But, to decouple a plugin is too much of a hassle and I have to stop scope-creep somewhere 😅 .

louis-bompart: (will rename)

francoislg (approved):
Great job, looks good!

I'd suggest waiting a bit before merging, I am currently trying to migrate to Jenkins and it doesn't currently publish new versions :thinking: 

Oh and you could bump the minor version right now!

louis-bompart:
👋 @dlafreniere can you run another pass so I can merge this :D

src/coveoua/plugins.ts
```diff
@@ -1,15 +1,48 @@
...
 export class Plugins {
-    private plugins: {[name: string]: any} = {};
+    public static readonly DefaultPlugins: string[] = [EC.Id, SVC.Id];
+    private registeredPluginsMap: Record<string, PluginClass> = {
+        [EC.Id]: EC,
+        [SVC.Id]: SVC,
+    };
+    private requiredPlugins: Record<string, BasePlugin> = {};
+
+    require(name: string, option: PluginOptions): void {
```
dlafreniere, line 16: `option` -> `options`

dlafreniere (approved):

package.json
```diff
@@ -39,14 +38,15 @@
...
-        "express": "^4.17.1",
         "fetch-mock": "^9.0.0-beta.2",
         "husky": "^4.3.0",
         "isomorphic-fetch": "2.2.1",
         "jest": "^25.1.0",
         "jsdom": "^15.0.0",
         "lint-staged": "^10.5.1",
         "prettier": "2.1.2",
+        "react": "^17.0.1",
+        "react-native": "^0.63.3",
```
tedre191, line 49: Why do we have a dep. to react and react-native?

tedre191 (approved):

package.json
louis-bompart: #226

PR #248, opened 2021-01-08
Try to checkout branch and run snyk using Jenkins DSL

francoislg:
[COM-765]

One more! 

The snyk results were not properly included in the deployment package, so I am attempting `runSnyk` in hope that it will work better.

Also, the version bump did not work, I had this error:

```
13:58:35  fatal: You are not currently on a branch.
13:58:35  To push the history leading to the current (detached HEAD)
13:58:35  state now, use
13:58:35  
13:58:35      git push origin HEAD:<name-of-remote-branch>
```

So I included a `git checkout master`

[COM-765]: https://coveord.atlassian.net/browse/COM-765

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -65,6 +65,7 @@ pipeline {
       steps {
         script {
           gitUtils.withCredentialHelper() {
+            sh "git checkout master"
             sh "npm version -m \"Bump version to %s [version bump]\""
           }
         }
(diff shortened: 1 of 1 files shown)


.deployment.config.json
```diff
@@ -24,7 +24,7 @@
         "no_container_images": true,
         "configurations": [
             {
-                "project_name": "coveo.analytics.js"
+                "project_name": "coveo-analytics-js"
```
Spuffynism, line 27: This doesn't need to match the github project?

Spuffynism (approved):

.deployment.config.json
francoislg: huh, might very well be, never occurred to me :smile: 

Will validate, I thought it only needed to match the project used to run the snyk script :thinking:

francoislg: I think you are right, so I updated it

cnguyen-coveo (approved):

PR #249, opened 2021-01-11
Put real dependency in package.json dependencies

francoislg:
[COM-765]

Snyk is currently failing because there was no `dependencies`, so I went through our code, and this was the only real thing that we included. (which is pretty good for this kind of lib, IMO)

[COM-765]: https://coveord.atlassian.net/browse/COM-765

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.16.0",
+    "version": "2.16.1",
     "lockfileVersion": 1,
     "requires": true,
     "dependencies": {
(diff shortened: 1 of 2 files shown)


toussaints (approved):

cnguyen-coveo (approved):

Spuffynism (approved):

dlafreniere (approved):
out of curiosity, how was the lib working if the dep wasn't specified?

francoislg:
> out of curiosity, how was the lib working if the dep wasn't specified?

Since we are building the package with Rollup, the dependency is included in the end-result. 

The only difference between `devDependencies` and `dependencies` is that `devDependencies` won't be installed if you `npm install coveo.analytics`. So I think there is no difference in the case, really...

But since Snyk wants to scan the included code in the lib, it makes sense to make it scan it :)

PR #258, opened 2021-02-04
add preprocessRequest to client [KIT-405]

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-405 (used here https://github.com/coveo/ui-kit/pull/483)

We have this feature in the headless platform client, to allow to override request parameters such as headers, method, url, etc.

Currently coveo.analytics only allowed us to add hooks to change the events payload, but there was nothing to modify the request itself before sending it.

We currently have a customer who wants to remove an authorization header to proxy through their platform without hiccups.

### DIFF
--- a/src/client/analytics.spec.ts
+++ b/src/client/analytics.spec.ts
@@ -373,7 +373,7 @@ describe('Analytics', () => {
     });
 
     it('should preprocess the request with the preprocessRequest', async () => {
-        let clientType: string;
+        let clientOrigin: string;
         let processedRequest: IAnalyticsRequestOptions = {
             url: 'https://www.myownanalytics.com/endpoint',
             headers: {
@@ -391,7 +391,7 @@ describe('Analytics', () => {
             endpoint: anEndpoint,
             version: A_VERSION,
             preprocessRequest: (request, type) => {
-                clientType = type;
+                clientOrigin = type;
                 processedRequest = {
                     ...request,
                     ...processedRequest,
@@ -405,7 +405,7 @@ describe('Analytics', () => {
         const options: RequestInit = call[1];
 
         const {url: expectedUrl, ...expectedOptions} = processedRequest;
-        expect(clientType).toBe('fetch');
+        expect(clientOrigin).toBe('analyticsFetch');
         expect(url).toBe(expectedUrl);
         expect(options).toEqual(expectedOptions);
     });
--- a/src/client/analytics.spec.ts
+++ b/src/client/analytics.spec.ts
@@ -1,6 +1,7 @@
 import * as fetchMock from 'fetch-mock';
 import {EventType, ViewEventRequest, DefaultEventResponse} from '../events';
 import {CoveoAnalyticsClient} from './analytics';
+import {IAnalyticsFetchClientCallOptions} from './analyticsFetchClient';
 import {CookieStorage} from '../storage';
 import HistoryStore from '../history';
 
@@ -369,6 +370,43 @@ describe('Analytics', () => {
             beforeSendHooks: [spy],
         }).sendEvent(EventType.search, searchEventPayload);
         expect(spy).toHaveBeenLastCalledWith(EventType.search, expect.objectContaining(searchEventPayload));
+    });
+
+    it('should preprocess the request with the preprocessRequestMiddleware', async () => {
+        const processedRequest: IAnalyticsFetchClientCallOptions = {
+            url: 'https://www.myownanalytics.com/endpoint',
+            credentials: 'same-origin',
+            headers: {
+                Age: '24',
+            },
+            method: 'PUT',
+            mode: 'same-origin',
+            payload: {
+                test: 'custom',
+            },
+        };
+        fetchMock.put(processedRequest.url, eventResponse);
+        const searchEventPayload = {queryText: 'potato'};
+        await new CoveoAnalyticsClient({
+            token: aToken,
+            endpoint: anEndpoint,
+            version: A_VERSION,
+            preprocessRequestMiddleware: () => {
+                return processedRequest;
+            },
+        }).sendEvent(EventType.search, searchEventPayload);
+
+        const call = fetchMock.calls()[0];
+        const url = call[0];
+        const options: RequestInit = call[1];
+        const body = getParsedBodyCalls()[0];
+
+        expect(url).toBe(processedRequest.url);
+        expect(options.credentials).toBe(processedRequest.credentials);
+        expect(options.headers).toEqual(processedRequest.headers);
+        expect(options.method).toBe(processedRequest.method);
+        expect(options.mode).toBe(processedRequest.mode);
+        expect(body).toEqual(processedRequest.payload);
     });
 
     const getParsedBodyCalls = (): any[] => {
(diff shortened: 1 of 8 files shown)


src/client/analyticsFetchClient.ts
```diff
@@ -1,26 +1,54 @@
...
+export interface IAnalyticsFetchClientCallOptions {
+    url: string;
+    headers: Record<string, string>;
+    method: HttpMethods;
+    mode: RequestMode;
+    credentials: RequestCredentials;
+    payload: Record<string, any>;
+}
+
+export type PreprocessAnalyticsRequestMiddleware = (
```
dlafreniere, line 15: its a bit of a mouthful....

src/client/analyticsFetchClient.ts
```diff
@@ -1,26 +1,54 @@
...
+    request: IAnalyticsFetchClientCallOptions
+) => IAnalyticsFetchClientCallOptions | Promise<IAnalyticsFetchClientCallOptions>;
+
 export interface IAnalyticsFetchClientOptions {
     baseUrl: string;
     token?: string;
     visitorIdProvider: VisitorIdProvider;
+    preprocessRequestMiddleware?: PreprocessAnalyticsRequestMiddleware;
```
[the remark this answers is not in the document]
ThibodeauJF: Good point, I'll adapt it for the beacon as well

src/client/analytics.ts
```diff
@@ -45,6 +45,7 @@ export interface ClientOptions {
     version: string;
     runtimeEnvironment?: IRuntimeEnvironment;
     beforeSendHooks: AnalyticsClientSendEventHook[];
+    preprocessRequestMiddleware?: PreprocessRequestMiddleware;
```
samisayegh, line 48: I think we could shorten the name to `preprocessRequest` without losing any clarity on what the option is supposed to do.

src/client/analytics.spec.ts
```diff
@@ -371,6 +372,42 @@ describe('Analytics', () => {
...
+        const call = fetchMock.calls()[0];
+        const url = call[0];
+        const options: RequestInit = call[1];
+
+        expect(url).toBe(processedRequest.url);
+        expect(options.credentials).toBe(processedRequest.credentials);
+        expect(options.headers).toEqual(processedRequest.headers);
+        expect(options.method).toBe(processedRequest.method);
+        expect(options.mode).toBe(processedRequest.mode);
+        expect(options.body).toBe(processedRequest.body);
```
samisayegh, line 408: Could we reduce the number of `expect` statements by comparing the entire `options` object? The test will also catch if a new property is added to the options. e.g.

```
const {url: expectedUrl, ...expectedOptions} = processedRequest
expect(url).toBe(expectedUrl)
expect(options).toBe(expectedOptions)
```

samisayegh (approved):

src/client/analyticsBeaconClient.ts
```diff
@@ -17,18 +23,23 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
...
+
         // tslint:disable-next-line: no-console
         console.log(`Sending beacon for "${eventType}" with: `, JSON.stringify(payload));
-        navigator.sendBeacon(
-            url,
-            new Blob([parsedRequestData], {
-                type: 'application/x-www-form-urlencoded',
-            })
-        );
+        navigator.sendBeacon(url, body as any); // https://github.com/microsoft/TypeScript/issues/38715
```
ThibodeauJF, line 42: Not setting to any returned this TS error:
```
[!] (plugin rpt2) Error: /Users/jfthibodeau/dev/coveo.analytics.js/src/client/analyticsBeaconClient.ts(42,35): semantic error TS2345: Argument of type 'string | ArrayBufferView | ArrayBuffer | Blob | FormData | URLSearchParams | ReadableStream<Uint8Array> | null | undefined' is not assignable to parameter of type 'string | ArrayBuffer | Blob | FormData | Uint8Array | Int8Array | Int16Array | Int32Array | Uint16Array | ... 6 more ... | undefined'.
  Type 'ArrayBufferView' is not assignable to type 'string | ArrayBuffer | Blob | FormData | Uint8Array | Int8Array | Int16Array | Int32Array | Uint16Array | ... 6 more ... | undefined'.
    Type 'ArrayBufferView' is missing the following properties from type 'DataView': getFloat32, getFloat64, getInt8, getInt16, and 13 more.
```

nrawji:
This is kind of tangentially related to this PR but one issue that users may encounter, is that if they don't provide a runtime, and they're in basically any modern browser, it'll default to `BrowserRuntime`. This uses the`AnalyticsBeaconClient`, which as you mentioned won't support things like headers. But they need to send a header, so they need to provide a custom runtime, which means they'll need create a custom client like follows:

```
import {NodeJSRuntime, CoveoAnalyticsClient, CookieAndLocalStorage} from 'coveo.analytics.js';

const client = new CoveoAnalyticsClient({
  runtimeEnviornment = new NodeJSRuntime(new CookieAndLocalStorage()));
});
```
And then there are more edge cases like when LocalStorage isn't available or they don't want to use cookies. So basically the point I'm trying to make is there's no easy way to use `BrowserStorage` with the `AnalyticsFetchClient` and that might limit the flexibility end users have in being able to pre-process the request. I'm not sure that this will actually be a huge use case for users (the only real example I can think of is adding a header if they proxy requests through their own authenticated endpoint) but it is pretty unintuitive for end users right now.

nrawji (approved):

src/client/analyticsRequestClient.ts
```diff
@@ -8,3 +8,11 @@ export interface VisitorIdProvider {
...
     sendEvent(eventType: string, payload: IRequestPayload): Promise<AnyEventResponse | void>;
 }
+
+export type PreprocessRequest = (
+    request: IAnalyticsRequestOptions
+) => IAnalyticsRequestOptions | Promise<IAnalyticsRequestOptions>;
+
+export interface IAnalyticsRequestOptions extends RequestInit {
+    url: string;
+}
```
francoislg, line 18: Would it be useful to expose the "origin" of the request? Like `clientOrigin: 'fetch' | 'beacon'`, to make conditional changes very convenient depending on the client?

Something like:

```
export type PreprocessRequest = (
    request: IAnalyticsRequestOptions,
    clientOrigin: 'fetch' | 'beacon'
) => IAnalyticsRequestOptions | Promise<IAnalyticsRequestOptions>;
```

Maybe it could even be strongly typed? AKA the second parameter is `beacon`, then the request can only have `url + body` instead of the whole RequestInit? :thinking:  Just brainstorming here :smile:

src/client/analyticsRequestClient.ts
```diff
@@ -8,3 +8,11 @@ export interface VisitorIdProvider {
 export interface AnalyticsRequestClient {
     sendEvent(eventType: string, payload: IRequestPayload): Promise<AnyEventResponse | void>;
 }
+
+export type PreprocessRequest = (
+    request: IAnalyticsRequestOptions
+) => IAnalyticsRequestOptions | Promise<IAnalyticsRequestOptions>;
+
+export interface IAnalyticsRequestOptions extends RequestInit {
```
dlafreniere, line 16: I'm not 100% sure, but I think if you use `Request`, it already has the `url: string` member in it

PR #260, opened 2021-02-16
Add support for language in SearchPageClient

olamothe:
Need a way to specify the language if not available from the `addDefaultValues` function.

For example, when running in an environment different from a browser, `document.documentElement.lang` will never exists.

In a browser scenario, this will be null/undefined if the end users did not set `<html lang="foo">` in their markup.

Ref: https://coveord.atlassian.net/browse/KIT-447

### DIFF
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -262,7 +262,6 @@ export class CoveoSearchPageClient {
         const payload: ClickEventRequest = {
             ...info,
             ...this.getBaseEventRequest({...identifier, ...metadata}),
-            ...this.getOrigins(),
             searchQueryUid: this.provider.getSearchUID(),
             queryPipeline: this.provider.getPipeline(),
             actionCause: event,
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -32,6 +32,7 @@ export interface SearchPageClientProvider {
     getOriginLevel1: () => string;
     getOriginLevel2: () => string;
     getOriginLevel3: () => string;
+    getLanguage: () => string;
 }
 
 export interface SearchPageClientOptions extends ClientOptions {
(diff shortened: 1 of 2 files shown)


src/searchPage/searchPageClient.ts
```diff
@@ -262,19 +259,13 @@ export class CoveoSearchPageClient {
...
-        const customData = {
-            ...this.provider.getBaseMetadata(),
-            ...identifier,
-            ...metadata,
-        };
-
         const payload: ClickEventRequest = {
             ...info,
+            ...this.getBaseEventRequest({...identifier, ...metadata}),
             ...this.getOrigins(),
```
francoislg, line 265: I think you don't need this `getOrigins` since it is now included in `getBaseEventRequest` :thinking:

francoislg (approved):
Looks good

src/searchPage/searchPageClient.ts
olamothe: Indeed, good catch !

src/searchPage/searchPageClient.ts
```diff
@@ -294,18 +284,31 @@ export class CoveoSearchPageClient {
...
     }
 
+    private getBaseCustomEventRequest(metadata?: Record<string, any>) {
+        return {
+            ...this.getBaseEventRequest(metadata),
+            lastSearchQueryUid: this.provider.getSearchUID(),
+        };
+    }
+
+    private getBaseEventRequest(metadata?: Record<string, any>) {
```
samisayegh, line 303: Is there a reason to use `EventRequest` in this and the above function's name? If not, I would suggest `getCustomMetadata` and `getBaseMetadata`

samisayegh (approved):

src/searchPage/searchPageClient.ts
olamothe: It's more than just "metadata/custom data" in the analytics lingo (the map of arbitrary key values you can send with any UA events), but also the languages and origins (1/2/3)

PR #261, opened 2021-02-17
feat(it): test validating a request is sent on page load

francoislg:
[COM-800]

This here adds a simple HTML page served by the lightweight node package `http-server`, and a Cypress test that validates that the outgoing request and response are correct.

I think we can do more than that, but I felt this was a pretty good start :smile: 

We already have a Jenkins job running that should notify us if the thing is failing.

I tried to design the thing to allow logging to any environment, and most importantly, try any version. So in the long run, we could use this as a certifier in the deployment pipeline with a deployed `MAJOR_MINOR_PATCH` version before deploying to the `MAJOR`!

[COM-800]: https://coveord.atlassian.net/browse/COM-800

### DIFF
--- a/cypress/plugins/index.js
+++ b/cypress/plugins/index.js
@@ -1,0 +1,21 @@
+/// <reference types="cypress" />
+// ***********************************************************
+// This example plugins/index.js can be used to load plugins
+//
+// You can change the location of this file or turn off loading
+// the plugins file with the 'pluginsFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/plugins-guide
+// ***********************************************************
+
+// This function is called when a project is opened or re-opened (e.g. due to
+// the project's config changing)
+
+/**
+ * @type {Cypress.PluginConfig}
+ */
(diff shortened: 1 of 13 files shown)


cypress/plugins/index.js
```diff
@@ -0,0 +1,21 @@
+/// <reference types="cypress" />
```
Spuffynism, line 1: Do we need this file?

Spuffynism (approved):
I'm **so** glad we have this. Nice work! :rocket:

cypress/plugins/index.js
francoislg: The files are needed for Cypress, even if empty, yeah

Spuffynism: Huh, alright then! something something green boat

Spuffynism (approved):

louis-bompart (approved):
Looks good :D

toussaints (approved):

cnguyen-coveo (approved):

PR #264, opened 2021-03-23
SPAAS-647 Only minify browser bundle; remove sourcemaps from non-minifed bundles

samisayegh:
**Background**

I am trying to bundle the coveo.analytics package inside IPX. However, I do not want Chrome to show this warning related to source maps:

source-map_warning

**Problem**

By default, webpack tries to pull the file specified in the package.json `browser` field. This file should be minified and have a source map, since it is also served by CDN directly into browsers.

It is possible to configure webpack to use a different field such as `main` though. However, the file specified in the coveo.analytics `main` is also minified. This is not typical among open-source libs.

As far as I know, the best practice is to not minify non-browser files because they will be consumed by a bundler. This leaves it up to the consumer as to whether they want to minify the file and/or create source maps.

**Solution**

I adjusted rollup to no longer minify libUMD. I also removed source maps for this output, but also the `module` and `react-native` outputs for consistency.

Unrelated to the core issue, I adjusted the vscode configuration to show the `dist` folder, and removed a rollup plugin that was not being used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
         "rimraf": "^3.0.2",
         "rollup": "^2.7.3",
         "rollup-plugin-serve": "^1.0.1",
-        "rollup-plugin-terser": "^5.3.0",
         "rollup-plugin-typescript2": "^0.27.0",
         "rollup-plugin-uglify": "^6.0.4",
         "ts-jest": "^25.2.0",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,7 +2,6 @@
     "typescript.tsdk": "node_modules\\typescript\\lib",
     "files.exclude": {
         ".nyc_output": true,
-        "dist": true,
         "dist_test": true
     },
     "typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false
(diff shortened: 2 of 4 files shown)


dlafreniere (approved):

package.json
```diff
@@ -9,7 +9,7 @@
     "scripts": {
         "lint:check": "prettier --check .",
         "lint:fix": "prettier --write .",
-        "build": "rollup -c",
+        "build": "npm run clean && rollup -c",
```
francoislg, line 12: Why was this needed? I'd rather not delete anything on "build"

francoislg (approved):

package.json
samisayegh: I thought my changes to rollup were not working because I could still see the `.js.map` source map files in the `dist` folder. I then realized that this was because the directory was not emptied from the files generated by the last run before the changes.

I feel cleaning before each build helps avoid such scenarios, but if you prefer to not do this, I will revert this change.

francoislg: `clean` cleans up a bit more things than the build, which is the main reason I am not super fond of it :sweat: 

I think we rarely **need** a clean state so the implicit delete does not look required to me

samisayegh: Ok, I will revert the change 👍

PR #265, opened 2021-03-24
Extend mappings with coveo specific keys

francoislg:
This adds the new keys that should be accepted by our protocol.

I also noticed a small thing in our implementation.

The following code:

```
coveoua("ec:setAction", "detail", {
  affiliation: 'x'
});
```

would map `affiliation` to `ta` (transaction affiliation) because we were using a somewhat global key translation.

So with our new "review" and "quote", it would need to be translated to *another key* than `ta`. We had the same issue with `id`, which shares its key with many actions, but not their mapping.

So I went into the [Measurement Protocol doc](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#(product_action)_affiliation) and noticed the following for the `Affiliation` attribute:

> The store or affiliation from which this transaction occurred. This is an additional parameter that can be sent when Product Action is set to 'purchase' or 'refund'.

This means that some attributes are *not* global. They should only be set when a specific action is set. 

So I implemented this logic with the `commerceActionKeysMappingPerAction`.

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -510,8 +510,8 @@ describe('ec events', () => {
                 ...defaultContextValues,
                 pa: 'quote',
                 t: 'event',
-                id: 'something',
-                affiliation: 'my super store',
+                quoteId: 'something',
+                quoteAffiliation: 'my super store',
             });
         });
 
@@ -526,7 +526,7 @@ describe('ec events', () => {
                 ...defaultContextValues,
                 pa: 'quote',
                 t: 'event',
-                affiliation: 'super store',
+                quoteAffiliation: 'super store',
             });
         });
 
@@ -543,7 +543,7 @@ describe('ec events', () => {
                 ...defaultContextValues,
                 pa: 'review',
                 t: 'event',
-                id: 'something',
+                reviewId: 'something',
                 reviewRating: 5,
             });
         });
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -496,6 +496,76 @@ describe('ec events', () => {
         });
     });
 
+    describe('with the coveo extensions', () => {
+        it('can send a quote event with a specific id and affiliation', async () => {
+            await coveoua('ec:setAction', 'quote', {
+                id: 'something',
+                affiliation: 'my super store',
+            });
+            await coveoua('send', 'event');
+
+            const [body] = getParsedBody();
+
+            expect(body).toEqual({
+                ...defaultContextValues,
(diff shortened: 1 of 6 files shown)


functional/ec-events.spec.ts
```diff
@@ -496,6 +496,76 @@ describe('ec events', () => {
         });
     });
 
+    describe('with the coveo extensions', () => {
+        it('can send a quote event with a specific id and affiliation', async () => {
```
toussaints, line 500: Should we test that we _cannot_ set some attributes on events? Like affiliation in a review action?
What do we expect? The attribute to be filtered out?

functional/ec-events.spec.ts
```diff
@@ -496,6 +496,76 @@ describe('ec events', () => {
...
+                t: 'event',
+                id: 'something',
+                reviewRating: 5,
+            });
+        });
+
+        it('can send a product with the group property', async () => {
+            await coveoua('ec:addProduct', {
+                id: 'something',
+                group: 'nsync',
```
toussaints, line 554: https://www.youtube.com/watch?v=Eo-KmOd3i7s

src/client/measurementProtocolMapping/commerceMeasurementProtocolMapper.ts
```diff
@@ -41,9 +43,53 @@ const transactionActionsKeysMapping: {[name: string]: string} = {
...
+        ...transactionActionsKeysMapping,
+    },
+    quote: {
+        ...productActionsKeysMapping,
+        ...quoteActionsKeysMapping,
+    },
+    review: {
+        ...productActionsKeysMapping,
+        ...reviewActionsKeysMapping,
+    },
```
toussaints, line 92: I like that. Pretty clear to me

toussaints (approved):

dlafreniere (approved):
nice find, clean solution

functional/ec-events.spec.ts
```diff
@@ -496,6 +496,76 @@ describe('ec events', () => {
...
+            expect(body).toEqual({
+                ...defaultContextValues,
+                pa: 'quote',
+                t: 'event',
+                id: 'something',
+                affiliation: 'my super store',
+            });
+        });
+
+        it('can send set an affiliation before defining the quote action', async () => {
```
Spuffynism, line 518:
```suggestion
        it('can set an affiliation before defining the quote action', async () => {
```

Spuffynism (approved):

ssharmaCoveo (approved):

PR #271, opened 2021-03-30
feat: expose handleOneAnalyticsEvent to headless

lbergeron:
https://coveord.atlassian.net/browse/SFINT-3742

In order to log the Usage Analytics events for Service from headless, we need to get access to the `handleOneAnalyticsEvent` function.

This PR, makes the function accessible.

### DIFF
--- a/src/coveoua/headless.ts
+++ b/src/coveoua/headless.ts
@@ -1,4 +1,5 @@
 export {CoveoSearchPageClient, SearchPageClientProvider} from '../searchPage/searchPageClient';
 export {CoveoAnalyticsClient, AnalyticsClientSendEventHook} from '../client/analytics';
 export {PreprocessAnalyticsRequest} from '../client/analyticsRequestClient';
+export {handleOneAnalyticsEvent} from './simpleanalytics';
 export * as history from '../history';


francoislg (commented):
:thinking: `handleOneAnalyticsEvent` is mostly used for the `coveoua` variable and map the actions from the queue to actual function. In your case, I think using that makes you lose all the advantages of TypeScript since the call is very loosely typed.

For headless, I think that you already know the series of events that you want to provide, you don't need a queue nor easy-to-parse string commands, so I would think that creating a `CoveoAnalyticsClient` would be good enough.

Now, I think the issue is the plugin, right? Can we expose them instead?

You would then have:

```
const client = new CoveoAnalyticsClient({
   ...
});
servicePlugin.setAction("name", {...});
servicePlugin.setTicket({...});
client.send("collect", {...});
```

with real types validation.

This is pretty much how the JSUI does it (without the plugins), and I think it will be easier to maintain in the long run :thinking:

lbergeron:
@francoislg Thanks for your input.

> Now, I think the issue is the plugin, right? Can we expose them instead?

That is indeed the issue. Exposing `handleOneAnalyticsEvent` was an easy way to get what I needed. But yes, I think you are right that exposing the plugins would be cleaner. I'll look into that. Thanks

francoislg:
@lbergeron I'd be OK with merging this if you just want to get the feature out and improve it afterward by using the clients :smile: 

Your pick!

lbergeron:
> @lbergeron I'd be OK with merging this if you just want to get the feature out and improve it afterward by using the clients 
> 
> Your pick!

@francoislg I explored the typed client approach, and finally I will keep the `handleOneAnalyticsEvent`.

Despite the fact that the function is loosely typed, on the headless side, the code (even more true in tests) is much more concise and simple.

francoislg (approved):

lbergeron:
After discussions [on this review](https://github.com/coveo/ui-kit/pull/660), it was decided that exposing `handleOneAnalyticsEvent` wasn't the right thing to do. `coveo.analytics.js` should instead expose a simplified and strongly-typed client to support Case Assist specific events.

PR #274, opened 2021-04-12
add getFacetState method to the SearchPageClientProvider

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-589

Facet state should be sent all the time and not just on the facet specific events.
This was previously added here https://github.com/coveo/coveo.analytics.js/pull/235 and is still only used by headless as far as we know. I could make the parameters optional and deprecate them, if we worry about potentially breaking other implementations.

### DIFF
--- a/src/searchPage/searchPageClient.spec.ts
+++ b/src/searchPage/searchPageClient.spec.ts
@@ -1,6 +1,6 @@
 import * as fetchMock from 'fetch-mock';
 import {CoveoSearchPageClient, SearchPageClientProvider} from './searchPageClient';
-import {SearchPageEvents, PartialDocumentInformation, CustomEventsTypes, FacetStateMetadata} from './searchPageEvents';
+import {SearchPageEvents, PartialDocumentInformation, CustomEventsTypes} from './searchPageEvents';
 import CoveoAnalyticsClient from '../client/analytics';
 import {NoopAnalytics} from '../client/noopAnalytics';
 
@@ -85,21 +85,7 @@ describe('SearchPageClient', () => {
             queryPipeline: 'my-pipeline',
             actionCause,
             customData,
-            language: 'en',
-            ...expectOrigins(),
-        });
-    };
-
-    const expectMatchFacetPayload = (actionCause: SearchPageEvents, meta = {}, facetState: FacetStateMetadata[]) => {
-        const [, {body}] = fetchMock.lastCall();
-        const customData = {foo: 'bar', ...meta};
-        expect(JSON.parse(body.toString())).toMatchObject({
-            queryText: 'queryText',
-            responseTime: 123,
-            queryPipeline: 'my-pipeline',
-            actionCause,
-            customData,
-            facetState,
+            facetState: fakeFacetState,
             language: 'en',
             ...expectOrigins(),
         });
@@ -293,7 +279,7 @@ describe('SearchPageClient', () => {
             facetTitle: 'title',
         };
         await client.logFacetSearch(meta);
-        expectMatchFacetPayload(SearchPageEvents.facetSearch, meta, fakeFacetState);
+        expectMatchPayload(SearchPageEvents.facetSearch, meta);
     });
 
     it('should send proper payload for #logFacetSelect', async () => {
@@ -305,7 +291,7 @@ describe('SearchPageClient', () => {
         };
 
         await client.logFacetSelect(meta);
-        expectMatchFacetPayload(SearchPageEvents.facetSelect, meta, fakeFacetState);
+        expectMatchPayload(SearchPageEvents.facetSelect, meta);
     });
 
     it('should send proper payload for #logFacetSelect', async () => {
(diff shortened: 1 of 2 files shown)


src/searchPage/searchPageClient.ts
```diff
@@ -270,17 +268,8 @@ export class CoveoSearchPageClient {
...
-        metadata: Record<string, any>,
-        facetState: FacetStateMetadata[]
-    ) {
-        const payload = {
-            ...this.getBaseSearchEventRequest(event, metadata),
-            facetState,
-        };
-
-        return this.coveoAnalyticsClient.sendSearchEvent(payload);
+    public logFacetSearchEvent(event: SearchPageEvents, metadata: Record<string, any>) {
```
olamothe, line 271: I think `logFacetSearchEvent` could be removed actually ? 

Since now "facet search" is no different/expect same payload as a "normal search"

src/searchPage/searchPageClient.ts
```diff
@@ -306,6 +295,7 @@ export class CoveoSearchPageClient {
             ...this.getOrigins(),
             customData,
             language: this.provider.getLanguage(),
+            facetState: this.provider.getFacetState ? this.provider.getFacetState() : [],
```
olamothe, line 298: Do we need to add the facet state to all events (custom/click/search) ? 

Or only search ?

Just make sure that it's added to the correct "buildRequest" function.

src/searchPage/searchPageClient.spec.ts
```diff
@@ -45,7 +39,7 @@ describe('SearchPageClient', () => {
 
     let client: CoveoSearchPageClient;
 
-    const provider = {
+    const provider: SearchPageClientProvider = {
```
olamothe, line 42: You should probably modify the UT in this file to get a better coverage of what is actually expected by ML.

`expectMatchFacetPayload` should probably be removed and merged with `expectMatchPayload`.

Otherwise, we are only checking the availability of facet state in the payload for facet events events, which is not complete. 

Is the exact thing we are trying to fix with this PR after all.

olamothe (approved):

src/searchPage/searchPageClient.ts
```diff
@@ -102,11 +103,8 @@ export class CoveoSearchPageClient {
         return this.logSearchEvent(SearchPageEvents.searchboxAsYouType);
     }
 
-    public logBreadcrumbFacet(
-        metadata: FacetMetadata | FacetRangeMetadata | CategoryFacetMetadata,
-        facetState: FacetStateMetadata[]
-    ) {
```
francoislg, line 108: This is sadly a breaking change :/

Are you sure only the JSUI team is using this?

I'd be ready to accept the change if your team feels comfortable with it

src/searchPage/searchPageClient.ts
```diff
@@ -33,6 +33,7 @@ export interface SearchPageClientProvider {
     getOriginLevel2: () => string;
     getOriginLevel3: () => string;
     getLanguage: () => string;
+    getFacetState?: () => FacetStateMetadata[];
```
samisayegh, line 36: Why is the new method optional? Shouldn't all classes implementing the `SearchPageClientProvider` interface implement the method?

samisayegh (approved):

src/searchPage/searchPageClient.ts
ThibodeauJF: It would be a breaking change for anyone upgrading

samisayegh: I see, I think it would be better for the code-base long-term to create a dedicated interface that extends `SearchPageClientProvider`, but I will leave @francoislg the final word,

ThibodeauJF: If we extend it, the original users of SearchPageClientProvider would have to use another interface if they want to use the facet state? Facets will often not be used by customers, having it optional kind of makes sense

src/searchPage/searchPageClient.ts
ThibodeauJF: Yes, asked ML and it's a plus if we add it for all types of events 😃

src/searchPage/searchPageClient.ts
ThibodeauJF: We added these recently for Headless, I'm pretty sure that my team agrees with removing them, but as I said in the description, we can play it safe and keep them as optional.
@olamothe  @samisayegh  @btaillon  ?

samisayegh: It's likely no one is using it, but it would be safer to deprecate the param and mark it as optional imo. Before v3 is released, one could go through and clear all the deprecated things,

ThibodeauJF: Discussed with the team this morning and we are ready to go forward with the change

francoislg (approved):
Looks good

PR #275, opened 2021-04-20
fix(ec): ensure that unknown actions work properly

francoislg:
+ add impression as first-class action

[COM-986]

The `measurementProtocolMapper` was the actual issue, if the action was not supported, it would throw at `const newKey = keysMappingForAction[key] || measurementProtocolKeysMapping[key] || key;`. It is now tested with `doesn't crash when sending an unknown action`.

The `impression` line is to properly map the `impression` action. Without that, the `can send a product impression event` was failing because it didn't set the `pa` attribute

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -115,6 +115,23 @@ describe('ec events', () => {
         });
     });
 
+    it('can send a product impression event', async () => {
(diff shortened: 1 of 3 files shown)


cnguyen-coveo (approved):

Spuffynism (approved):

francoislg:
screenshot-2021-04-20-09-40-58

Here is proof the test validates the correct one :P

dlafreniere (approved):

PR #277, opened 2021-04-22
[COM-971] add getParameters and getPayload to analyticsClient

cnguyen-coveo:
Big shoutout to **Frank** for all the help! 

Split the `sendEvent()` to be able to get both parameters (non altered info) and payload.
From the browser, if you run `window.coveoanalytics.getCurrentClient().getParameters('event', {});` you should get an output.

Screen Shot 2021-04-22 at 4 12 44 PM

Screen Shot 2021-04-22 at 4 13 21 PM

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -1,7 +1,11 @@
 import 'isomorphic-fetch';
 import * as fetchMock from 'fetch-mock';
 import {DefaultEventResponse} from '../src/events';
+import type {getCurrentClient} from "../src/coveoua/library";
 import coveoua from '../src/coveoua/browser';
+
+declare const self: any;
+const getClient: typeof getCurrentClient = (self.coveoanalytics as any).getCurrentClient;
 
 describe('ec events', () => {
     const initialLocation = `${window.location}`;
@@ -28,6 +32,8 @@ describe('ec events', () => {
         z: expect.stringMatching(guidFormat),
     };
 
+    let client: ReturnType<typeof getClient>;
+
     beforeEach(() => {
         changeDocumentLocation(initialLocation);
         const address = `${anEndpoint}/rest/v15/analytics/collect`;
@@ -42,6 +48,7 @@ describe('ec events', () => {
         });
         coveoua('reset');
         coveoua('init', aToken, anEndpoint);
+        client = getClient();
     });
 
     it('can set custom data at the root level and send a page view event', async () => {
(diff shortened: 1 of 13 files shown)


dlafreniere:
this is to help debug?

cnguyen-coveo:
> this is to help debug?

Yes, it's not really a bug but Luca is asking to have a way to expose the data which the methods would expose parameters ("raw" data) and payload ("transformed" data)

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+            },
+        });
+        const secondAfterPayload = await client.getPayload('pageview', {
+            title: 'wow',
+            custom: {
+                verycustom: 'value',
+            },
+        });
+
+        expect(firstPayload.t).toEqual(secondPayload.t);
```
francoislg, line 236: :thinking: I think you could use `expect(firstPayload).toEqual(secondPayload)` ?

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+        expect(firstAfterParameters.screenColor).toEqual(secondAfterParameters.screenColor);
+        expect(firstAfterParameters.screenResolution).toEqual(secondAfterParameters.screenResolution);
+        expect(firstAfterParameters.title).toEqual(secondAfterParameters.title);
+        expect(firstAfterParameters.userAgent).toEqual(secondAfterParameters.userAgent);
+        expect(firstAfterParameters.language).toEqual(secondAfterParameters.language);
+
+        expect(firstAfterParameters.location).toBe(secondLocation);
+    });
+
+    it('should return similar parameters and payload', async () => {
```
francoislg, line 320: I like this one, great idea :+1:

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+        expect(parameters.encoding).toEqual(payload.de);
+        expect(parameters.location).toEqual(payload.dl);
+        expect(parameters.referrer).toEqual(payload.dr);
+        expect(parameters.screenColor).toEqual(payload.sd);
+        expect(parameters.screenResolution).toEqual(payload.sr);
+        expect(parameters.title).toEqual(payload.dt);
+        expect(parameters.userAgent).toEqual(payload.ua);
+        expect(parameters.language).toEqual(payload.ul);
+    });
+
```
francoislg, line 336: I think I'd also add one test where you get the payload, send the event, and compare `const [event] = getParsedBody()` with the payload :thinking: 

This would ensure that the `getPayload` does return the proper "next" values :P

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+        expect(firstParameters.hitType).toEqual(secondParameters.hitType);
+        expect(firstParameters.pageViewId).toEqual(secondParameters.pageViewId);
+        expect(firstParameters.encoding).toEqual(secondParameters.encoding);
+        expect(firstParameters.location).toEqual(secondParameters.location);
+        expect(firstParameters.referrer).toEqual(secondParameters.referrer);
+        expect(firstParameters.screenColor).toEqual(secondParameters.screenColor);
+        expect(firstParameters.screenResolution).toEqual(secondParameters.screenResolution);
+        expect(firstParameters.title).toEqual(secondParameters.title);
+        expect(firstParameters.userAgent).toEqual(secondParameters.userAgent);
+        expect(firstParameters.language).toEqual(secondParameters.language);
```
francoislg, line 302:
```suggestion
        expect(firstParameters).toEqual(secondParameters);
```

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+        expect(firstAfterPayload.t).toEqual(secondAfterPayload.t);
+        expect(firstAfterPayload.pid).toEqual(secondAfterPayload.pid);
+        expect(firstAfterPayload.de).toEqual(secondAfterPayload.de);
+        expect(firstAfterPayload.dl).toEqual(secondAfterPayload.dl);
+        expect(firstAfterPayload.dr).toEqual(secondAfterPayload.dr);
+        expect(firstAfterPayload.sd).toEqual(secondAfterPayload.sd);
+        expect(firstAfterPayload.sr).toEqual(secondAfterPayload.sr);
+        expect(firstAfterPayload.dt).toEqual(secondAfterPayload.dt);
+        expect(firstAfterPayload.ua).toEqual(secondAfterPayload.ua);
+        expect(firstAfterPayload.ul).toEqual(secondAfterPayload.ul);
```
francoislg, line 258:
```suggestion
        expect(firstAfterPayload).toEqual(secondAfterPayload);
```

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+        expect(firstAfterParameters.hitType).toEqual(secondAfterParameters.hitType);
+        expect(firstAfterParameters.pageViewId).toEqual(secondAfterParameters.pageViewId);
+        expect(firstAfterParameters.encoding).toEqual(secondAfterParameters.encoding);
+        expect(firstAfterParameters.location).toEqual(secondAfterParameters.location);
+        expect(firstAfterParameters.referrer).toEqual(secondAfterParameters.referrer);
+        expect(firstAfterParameters.screenColor).toEqual(secondAfterParameters.screenColor);
+        expect(firstAfterParameters.screenResolution).toEqual(secondAfterParameters.screenResolution);
+        expect(firstAfterParameters.title).toEqual(secondAfterParameters.title);
+        expect(firstAfterParameters.userAgent).toEqual(secondAfterParameters.userAgent);
+        expect(firstAfterParameters.language).toEqual(secondAfterParameters.language);
```
francoislg, line 315:
```suggestion
        expect(firstAfterParameters).toEqual(secondAfterParameters);
```

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,137 @@ describe('ec events', () => {
...
+        expect(parameters.hitType).toEqual(payload.t);
+        expect(parameters.pageViewId).toEqual(payload.pid);
+        expect(parameters.encoding).toEqual(payload.de);
+        expect(parameters.location).toEqual(payload.dl);
+        expect(parameters.referrer).toEqual(payload.dr);
+        expect(parameters.screenColor).toEqual(payload.sd);
+        expect(parameters.screenResolution).toEqual(payload.sr);
+        expect(parameters.title).toEqual(payload.dt);
+        expect(parameters.userAgent).toEqual(payload.ua);
+        expect(parameters.language).toEqual(payload.ul);
```
francoislg, line 334:
```suggestion
        expect(parameters).toEqual({
          hitType: payload.t,
          pageViewId: payload.pid,
          encoding: payload.de,
          location: payload.dl,
          referrer: payload.dr,
          screenColor: payload.sd,
          screenResolution: payload.sr,
          title: payload.dt,
          userAgent: : payload.ua,
          language: : payload.ul
        });
```

src/plugins/BasePlugin.ts
```diff
@@ -80,28 +81,42 @@ export abstract class BasePlugin {
...
+        this.currentLocation = location;
+
+        if (eventType === BasePluginEventTypes.pageview) {
+            this.nextPageViewId = this.uuidGenerator();
+        }
+
+        this.hasSentFirstPageView = true;
+    }
+
+    private getNextValues(eventType: string, payload: any) {
```
francoislg, line 105: I think we should add a comment in the code here to not lose track of *why* we did this.

Reformulate it how you want, but I'd put something like: 

> When calling `getPayload` or `getParameters`, we need to return what would be sent by the API without updating our internal reference. For instance, if you `getPayload("pageview")`, we want to return the same payload as if you did `coveoua("send", "pageview")`.

src/plugins/BasePlugin.ts
```diff
@@ -80,28 +81,42 @@ export abstract class BasePlugin {
...
+        return {
+            pageViewId: eventType === BasePluginEventTypes.pageview ? this.nextPageViewId : this.pageViewId,
```
francoislg, line 107: > When sending a pageview, we need to generate a new pageview ID.
> For instance, with the following sequence: 1. pageview, 2. event, 3. pageview, 4. event, 1 and 2 will have the pageviewid, whereas 3 and 4 will have a different pageviewid.
> We pre-generate the "nextPageViewId" in case you send the following sequence: 1. pageview, 2. event, 3. getPayload, 4. pageview. This ensures that 3 and 4 have the same pageviewid.
> We applied the same logic to the referrer and the location.

francoislg (approved):
Couple of comments, but nothing major :smile:

functional/ec-events.spec.ts
cnguyen-coveo: It doesnt work because the timestamp and eventId are different (which would be normal)

francoislg: :thinking: indeed, I'd still rather use an object equality thing, it checks more things than a couple of simple equals like that.

Let's figure out a way to ignore those two :smile:

dlafreniere (approved):

functional/ec-events.spec.ts
```diff
@@ -196,6 +203,197 @@ describe('ec events', () => {
...
+            },
+        });
+        const secondAfterPayload = await client.getPayload('pageview', {
+            title: 'wow',
+            custom: {
+                verycustom: 'value',
+            },
+        });
+
+        delete firstPayload.z;
```
francoislg, line 236: After putting those in a private method, all good

francoislg (approved):
:confetti_ball: amazing!

PR #278, opened 2021-04-30
SPAAS-716 Add SameSite=Lax when setting cookies

samisayegh:
There was a change to the cookie spec in the past year, adding the [SameSite](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite) attribute to better protect cookies from being used maliciously.

When the attribute is not set, browsers such as Firefox show the following warning in the console:

firefox_same-site_warning

It took me some time to figure out, but since the cookie is first-party (i.e associated with the customer's domain), I am 99% sure that `Lax` is the correct value, allowing scripts like IPX to read the value on first load.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
         "build": "rollup -c",
         "start": "rollup -c -w --environment SERVE",
         "test": "jest --clearCache && jest --coverage",
+        "test:watch": "jest --watch",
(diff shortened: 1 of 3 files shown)


francoislg (approved):
Looks good!

cnguyen-coveo (approved):

Spuffynism (approved):

dlafreniere (approved):

PR #279, opened 2021-05-21
Adding quickview, bookmark_add and bookmark_remove to supported product actions [COM-1079]

toussaints:
The `quickview` is being stripped out of the payload when using coveo.analytics.js leading to a 400 on the collect endpoint.

Turns out we were also missing `bookmark_add` and `bookmark_remove` from the list

See:
https://discuss.coveo.com/t/support-00068203-400-response-when-the-quick-view-event-is-fired/5855

They are supposed to be supported according to the doc
https://docs.coveo.com/en/l29e0540/coveo-for-commerce/tracking-commerce-events-reference#action-types-reference

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -1,7 +1,7 @@
 import 'isomorphic-fetch';
 import * as fetchMock from 'fetch-mock';
 import {DefaultEventResponse} from '../src/events';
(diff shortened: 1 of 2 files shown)


cnguyen-coveo (approved):

francoislg (approved):
Looks good, thanks for the new tests.

Spuffynism (approved):
:100:

PR #281, opened 2021-06-04
KIT-716 Run coveo.analytics on the server

samisayegh:
I was working on a [server-side rendered (SSR) PoC](https://github.com/coveo/ui-kit/pull/863) with headless this week and I noticed that the coveo.analytics library was throwing an error due to fetch not being available.

This PR makes the `coveo.analytics` package isomorphic, by creating dedicated bundles for node.js and web run-times. I experimented with the `isomorphic-fetch` npm package, but after not being able to bundle it correctly I switched to `cross-fetch`. This required me to slightly adjust the tests to mock `fetch` correctly since it was now being imported rather than accessed via a global.

I tested with headless on the server, and was able to log analytics successfully.


functional/ec-events.spec.ts
```diff
@@ -1,11 +1,12 @@
...
-import * as fetchMock from 'fetch-mock';
+import {sandbox} from 'fetch-mock';
 import {DefaultEventResponse} from '../src/events';
 import type {getCurrentClient} from '../src/coveoua/library';
 import coveoua from '../src/coveoua/browser';
+import * as CrossFetch from 'cross-fetch';
 
 declare const self: any;
 const getClient: typeof getCurrentClient = (self.coveoanalytics as any).getCurrentClient;
+const fetchMock = sandbox();
```
francoislg, line 9: I'm wondering if there could be a way to make it reusable across test :thinking: 

what about:

```
import {sandbox} from 'fetch-mock';
import * as CrossFetch from 'cross-fetch';

function mockFetch() {
   const fetchMock = sandbox();   
   return {
        fetchMock,
        fetchMockBeforeEach: () => jest.spyOn(CrossFetch, 'fetch').mockImplementation(fetchMock);
   }
}
```

that you could reuse with:

```
const {fetchMock, fetchMockBeforeEach} = mockFetch();

...

beforeEach(() => {
   fetchMockBeforeEach();
});
```

Still not sure about this, your solution is pretty simple enough, so ignore if you don't like it :)

francoislg (approved):
Looks good :+1:

bundle/browser-fetch.ts
```diff
@@ -0,0 +1,3 @@
+// This file is to prevent including a fetch polyfill from the "cross-fetch" dependency when building for a browser target.
+// Major browsers today provide fetch on the window object, so including a polyfill would needlessly increase the size of the bundle.
+export const fetch = window.fetch;
```
dlafreniere, line 3: do we have a list of support browsers? is there a way to make this degrade gracefully when `fetch` is not present?

dlafreniere (approved):
looks good, agree with Frank's comment for mock setup reuse

bundle/browser-fetch.ts
samisayegh: There is [this table](https://caniuse.com/?search=fetch) showing the browser versions. Most browsers provide `fetch` but the notable exception is IE11.

On degrading gracefully, we could ship a fetch polyfill with coveoua, though I wouldn't recommend it because it's pretty heavy. It would add ~15 kb minified to the bundle (the bundle goes from 45 -> 60 kb, so ~25% of the final size).

There is a check [here](https://github.com/coveo/coveo.analytics.js/blob/master/src/coveoua/browser.ts#L15) that logs an actionable message to the console when fetch is not available. Could that be enough?

functional/ec-events.spec.ts
samisayegh: Looks good, I will make these changes,

bundle/browser-fetch.ts
dlafreniere: perfect, didn't know that was already there!

francoislg: The log is enough! 

It was a conscious decision to remove the polyfill in the newer versions to reduce the payload :smile: 

If your site wants to support IE11 or other paleolithic browsers, include the polyfill yourself :P

PR #283, opened 2021-06-16
Add support for smart snippet analytics events

olamothe:
SmartSnippet (aka: Question answering).

We are adding support for the feature in headless.

### DIFF
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -20,6 +20,8 @@ import {
     FacetSortMeta,
     QueryErrorMeta,
     FacetStateMetadata,
+    SmartSnippetFeedbackReason,
+    SmartSnippetSuggestionMeta,
 } from './searchPageEvents';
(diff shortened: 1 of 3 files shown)


dlafreniere (approved):

src/searchPage/searchPageClient.ts
```diff
@@ -224,6 +226,42 @@ export class CoveoSearchPageClient {
         return this.logSearchEvent(SearchPageEvents.queryErrorClear);
     }
 
+    public logLikeSmartSnippet() {
```
francoislg, line 229: This is a `CoveoSearchPageClient` class, right?

So are those really related to "Search Page" events?

Should they instead be put in their own client?

olamothe: It's for classic search page, yeah.

They appear as part of a "normal" search request, like results list and facets.

francoislg (approved):

PR #285, opened 2021-06-17
[COM-160] notify slack when build goes back green

dlafreniere:
The build was failing, but then it wasn't obvious when it stopped failing

### DIFF
--- a/Jenkinsfile_FT
+++ b/Jenkinsfile_FT
@@ -76,9 +76,9 @@ pipeline {
   post {
     always {
       script {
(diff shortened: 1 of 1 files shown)


Spuffynism (approved):
I understand the logic change, but why the change from `currentResult` to `result`? :thinking:

dlafreniere:
> I understand the logic change, but why the change from `currentResult` to `result`? 🤔

for "consistency", since we are using `result` on the previousBuild object. In this case the two are equivalent because we are in the `post` block, so the result is guaranteed to be set.

basically `currentResult` is the "result so far", but here we are at the end of the build

cnguyen-coveo (approved):

toussaints (approved):

PR #287, opened 2021-07-26
COM-956 - fix(deploy): use real tgf directory structure

francoislg:
The job failed because it seems to be missing a "deployment" folder, so here it is


Spuffynism (approved):

dlafreniere:
The "new" directory structure actually does away with the module/deployment directories... but if it works 🤷

francoislg:
Following Devin's link, I will rework this thing

PR #291, opened 2021-07-29
add support for anonymous events

olamothe:
This tells Usage Analytics service to anonymize search/click/custom events.
From swagger:

> Whether the interaction that caused the search interface to log the event was triggered by an anonymous user. If set to true, the Usage Analytics Write API will not extract the name and userDisplayName, if present, from the search token

Need to add support for the parameter in headless

https://coveord.atlassian.net/browse/KIT-856

### DIFF
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -35,6 +35,7 @@ export interface SearchPageClientProvider {
     getOriginLevel2: () => string;
     getOriginLevel3: () => string;
     getLanguage: () => string;
+    getIsAnonymous: () => boolean;
     getFacetState?: () => FacetStateMetadata[];
 }
 
@@ -338,6 +339,7 @@ export class CoveoSearchPageClient {
             customData,
             language: this.provider.getLanguage(),
             facetState: this.provider.getFacetState ? this.provider.getFacetState() : [],
+            anonymous: this.provider.getIsAnonymous(),
         };
     }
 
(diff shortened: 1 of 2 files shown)


francoislg (approved):

dlafreniere (approved):

Spuffynism (approved):

src/searchPage/searchPageClient.ts
```diff
@@ -338,6 +339,7 @@ export class CoveoSearchPageClient {
             customData,
             language: this.provider.getLanguage(),
             facetState: this.provider.getFacetState ? this.provider.getFacetState() : [],
+            anonymous: this.provider.getIsAnonymous(),
```
samisayegh, line 342: Wounldn't it be simpler if a dev used an anonymous token from the get-go? Versus creating an authenticated token and overriding the behaviour of the analytics api client-side?

I'm curious why someone would want this combination? If it's a GDPR use-case, I think the law says no tracking at all, whether anonymous or not.

olamothe: It's something we support in the documentation: 

https://docs.coveo.com/en/2795/security/responding-to-data-subject-requests-with-coveo

GDPR stuff/requests that has been reviewed by legal in the past so... 🤷

samisayegh: Ok, I guess someone has a use-case where the search is authenticated but analytics are anonymous 🤷‍♂️

If feels weird, because if a person has exclusive access to e.g. their emails, and an analytics log shows a click on an email result, then we know that the person accessed their email via Coveo. The anonymous event is personally identifiable because the search needed to be authenticated. From there, one can use the visitorId to see what else they looked at ...

samisayegh (approved):

PR #306, opened 2021-10-04
[COM-1308] use doNotTrack and GPC to avoid setting cookies

dlafreniere:
doNotTrack was present but not used, I added support for GPC and tests.

If the browser indicates no tracking, instantiate a NoopRuntime


jebeaudet (approved):
Funny that having ublock on this page breaks everything because this lib is detected as ad 😂 

Thanks that was :sofast: you rock.

dlafreniere:
> Funny that having ublock on this page breaks everything because this lib is detected as ad 😂

yea, I had that too, had to add a filter
image

src/donottrack.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+        it('should respect GPC undefined', () => {
+            initModule(true, {
+                navigatorGlobalPrivacyControl: undefined,
+            });
+
+            expect(doNotTrack).toBeFalsy();
+        });
+
+        [true, 'yes', '1'].forEach((value) => {
+            it('should fallback `navigator.doNotTrack`: ' + JSON.stringify(value), () => {
```
Spuffynism, line 75: I don't think the `JSON.stringify` is necessary as values will be converted to strings automatically when being concatenated to one:
image

Spuffynism, line 75:
```suggestion
            it('should fallback on `navigator.doNotTrack`: ' + JSON.stringify(value), () => {
```

Spuffynism (approved):

src/client/analytics.spec.ts
```diff
@@ -55,6 +55,21 @@ describe('Analytics', () => {
         });
     });
 
+    it('should honor doNotTrack', () => {
+        jest.mock('../doNotTrack', () => ({
+            doNotTrack: false,
```
francoislg, line 60: :thinking: Shouldn't this be `true` for this test?

src/client/analytics.ts
```diff
@@ -131,7 +132,9 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
     }
 
     private initRuntime(clientsOptions: IAnalyticsBeaconClientOptions) {
-        if (hasWindow() && hasDocument()) {
+        if (doNotTrack) {
+            return new NoopRuntime();
```
francoislg, line 136: This is great! :+1:

src/donottrack.ts
```diff
@@ -6,10 +6,11 @@
 
 import {hasNavigator} from './detector';
 
-export const doNotTrack = hasNavigator()
-    ? [true, 'yes', '1'].indexOf(
+export const doNotTrack: boolean = hasNavigator()
+    ? !!(<any>navigator).globalPrivacyControl ||
+      [true, 'yes', '1'].indexOf(
           (<any>navigator).doNotTrack || (<any>navigator).msDoNotTrack || (<any>window).doNotTrack
```
francoislg, line 12: This might be a good opportunity to review this.

Are we sure this works as intended? For instance, if `navigator.doNotTrack = 'true'`, this will fail :P 

If `window.doNotTrack = true` but `navigator.doNotTrack = 'false'`, it will return `'false` because the string is a truthy value for the sake of `||` comparisons.

So the question is, should we not track if *any* of those values are set truthy?

If so, we could make it a simple `[(<any>navigator).globalPrivacyControl, (<any>navigator).doNotTrack,(<any>navigator).msDoNotTrack, (<any>window).doNotTrack].some(value => !!value)` instead

olamothe (approved):

src/donottrack.spec.ts
dlafreniere: your observation is correct, however I was explicitly using stringify to differentiate the booleans vs strings (because strings are quoted in json)

src/client/analytics.spec.ts
dlafreniere: you're right!

I think something is wrong with this test, I'll reevaluate

src/donottrack.ts
dlafreniere: each one has its own set of possible values... but I suppose checking them all for truthiness is safer

olamothe (commented):
I think one thing we should also do when there is a `doNotTrack` header is to erase any cookie/local storage currently present, which I think the current PR does not do.

It only prevents additional cookie being set.

PR #309, opened 2021-10-19
handle blocked local & session storage with a try/catch

ThibodeauJF:
accessing those throw errors when you disable access to cookies & other kinds of storage in the browser

https://coveord.atlassian.net/browse/KIT-1135

### DIFF
--- a/src/detector.ts
+++ b/src/detector.ts
@@ -11,11 +11,19 @@ export function hasDocument(): boolean {
 }
 
 export function hasLocalStorage(): boolean {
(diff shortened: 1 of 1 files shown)


ThibodeauJF:
I don't think it would change anything, we access the window correctly and the "localStorage" property is not defined, but Chrome` here will throw the following error right away when you try to access the property
Screen Shot 2021-10-19 at 4 26 57 PM

https://stackoverflow.com/questions/16427636/check-if-localstorage-is-available

PR #311, opened 2021-11-18
feat(case-assist): add Case Assist client

lbergeron:
https://coveord.atlassian.net/browse/SFINT-3835

This PR adds a strongly-typed client to log the Case Assist events. See [Log Case Assist Events](https://docs.coveo.com/en/3437/service/log-case-assist-events) for details regarding the information that are passed to the `svc` plugin.

### DIFF
--- a/src/caseAssist/caseAssistClient.ts
+++ b/src/caseAssist/caseAssistClient.ts
@@ -2,7 +2,10 @@ import CoveoAnalyticsClient, {AnalyticsClient, ClientOptions} from '../client/an
 import {NoopAnalytics} from '../client/noopAnalytics';
 import {SVCPlugin} from '../plugins/svc';
 import {
+    CaseAssistActions,
+    CaseAssistEvents,
     CaseCancelledMetadata,
+    CaseCancelledReasons,
     CaseCreatedMetadata,
     CaseSolvedMetadata,
     EnterInterfaceMetadata,
@@ -22,7 +25,7 @@ export class CaseAssistClient {
     private svc: SVCPlugin;
 
     constructor(private options: Partial<CaseAssistClientOptions>) {
-        this.client = options.enableAnalytics === false ? new NoopAnalytics() : new CoveoAnalyticsClient(options);
+        this.client = options.enableAnalytics ? new CoveoAnalyticsClient(options) : new NoopAnalytics();
         this.svc = new SVCPlugin({client: this.client});
     }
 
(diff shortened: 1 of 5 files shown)


src/caseAssist/caseAssistClient.ts
```diff
@@ -0,0 +1,112 @@
...
+        this.svc.setAction('ticket_cancel', {
+            reason: 'Quit',
+        });
+        this.svc.setTicket(meta.ticket);
+        return this.sendClickEvent();
+    }
+
+    public logCaseSolved(meta: CaseSolvedMetadata) {
+        this.svc.setAction('ticket_cancel', {
+            reason: 'Solved',
```
louis-bompart, line 93: Do you think the reason should be extracted into an `enum`?

src/caseAssist/caseAssistClient.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    public logSelectFieldSuggestion(meta: SelectFieldSuggestionMetadata) {
+        this.svc.setAction('ticket_classification_click', meta.suggestion);
+        this.svc.setTicket(meta.ticket);
+        return this.sendClickEvent();
+    }
+
+    public logSelectDocumentSuggestion(meta: SelectDocumentSuggestionMetadata) {
+        this.svc.setAction('suggestion_click', meta.suggestion);
```
louis-bompart, line 63: Do you think the action (e.g. `suggestion_click` etc) should be in an (or multiple) enums?

louis-bompart (approved):
lgtm overall 👍

src/caseAssist/caseAssistClient.ts
lbergeron: Yes I can add an enum for that.

src/caseAssist/caseAssistClient.ts
lbergeron: In fact, I should reuse the `CaseAssistActions` enum when setting the actions.

nathanlb (approved):
LGTM!

src/caseAssist/caseAssistClient.ts
```diff
@@ -0,0 +1,115 @@
...
+    EnterInterfaceMetadata,
+    MoveToNextCaseStepMetadata,
+    RateDocumentSuggestionMetadata,
+    SelectDocumentSuggestionMetadata,
+    SelectFieldSuggestionMetadata,
+    UpdateCaseFieldMetadata,
+} from './caseAssistActions';
+
+export interface CaseAssistClientOptions extends ClientOptions {
+    enableAnalytics: boolean;
```
francoislg, line 20: I guess this can be optional ?

francoislg (approved):
I didn't quite validate the format and everything, but the code itself fits well into what we have and it looks good :+1:

src/caseAssist/caseAssistClient.ts
lbergeron: I defined the option the same way as in `SearchPageClientOptions` to be consistent. But you are right that this flag could be optional, and should default to `true`.

lbergeron: @francoislg I just updated the PR to make `enableAnalytics` optional and default to `true`. Is that OK with you?

PR #312, opened 2021-11-23
fix: ensure the referrer is updated only on page view

francoislg:
[COM-1372]

[COM-1372]: https://coveord.atlassian.net/browse/COM-1372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -289,6 +289,31 @@ describe('ec events', () => {
         expect(afterSecond.dr).toBe(initialLocation);
     });
 
+    it('should only update the current location and referrer on the first page view (COM-1372)', async () => {
+        const secondLocation = 'http://very.new/';
+
+        await coveoua('send', 'event', '1');
+        await coveoua('send', 'pageview');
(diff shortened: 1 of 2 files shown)


dlafreniere (approved):

functional/ec-events.spec.ts
```diff
@@ -289,6 +289,31 @@ describe('ec events', () => {
         expect(afterSecond.dr).toBe(initialLocation);
     });
 
+    it('should only update the current location and referrer on the first page view (COM-1372)', async () => {
```
Spuffynism, line 292: Should we include the task number? We don't seem to usually do it in tests here.

Spuffynism (approved):
:+1: Nice, the test is super clear!

functional/ec-events.spec.ts
francoislg: We don't, but I thought it was nice :D

PR #314, opened 2021-12-06
DNT: add support for the new "unspecified" value

bigluck:
We're having some problems with the DNT support on Firefox.
The "unspecified" value used by firefox (and documentation here:  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/doNotTrack) is treated as a TRUE value by our library, enabling the Do Not Track for all the Firefox users even if they didn't enable it.

### DIFF
--- a/src/donottrack.spec.ts
+++ b/src/donottrack.spec.ts
@@ -94,5 +94,29 @@ describe('doNotTrack', () => {
                 expect(doNotTrack()).toBeTruthy();
             });
         });
+
+        [false, 'no', '0', 'unspecified'].forEach((value) => {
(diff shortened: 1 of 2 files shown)


src/donottrack.spec.ts
```diff
@@ -94,5 +94,29 @@ describe('doNotTrack', () => {
                 expect(doNotTrack()).toBeTruthy();
             });
         });
+
+        [false, 'no', '0', 'unspecified'].forEach((value) => {
```
francoislg, line 98: I like this!

francoislg (approved):
TOP

dlafreniere (approved):

fvanham-coveo (approved):

PR #321, opened 2022-02-10
Add clientId & originContext to searchPageClient events payload

ThibodeauJF:
https://coveord.atlassian.net/browse/KIT-1409

2 parameters that were sent with the JSUI & that are still relevant with Headless. The clientId is the same as the visitorId.

### DIFF
--- a/src/searchPage/searchPageClient.spec.ts
+++ b/src/searchPage/searchPageClient.spec.ts
@@ -56,6 +56,7 @@ describe('SearchPageClient', () => {
         }),
         getSearchUID: () => 'my-uid',
         getPipeline: () => 'my-pipeline',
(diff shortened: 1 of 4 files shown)


src/searchPage/searchPageClient.spec.ts
```diff
@@ -68,9 +69,9 @@ describe('SearchPageClient', () => {
         fetchMockBeforeEach();
 
         client = initClient();
+        client.coveoAnalyticsClient.runtime.storage.setItem('visitorId', 'visitor-id');
```
ThibodeauJF, line 72: in the tests, the interfaceLoad (first test) appear to be assigned a new uuid before the mock API response which sets the correct visitorId

btaillon (approved):

PR #326, opened 2022-02-15
builder pattern for search page events which allows to resurface event action cause and metadata

olamothe:
This is the accompanying PR with the analytics action cause and custom data change in Headless: 
https://github.com/coveo/ui-kit/pull/1738

It's a lot of change, but the changes are very straightforward and repetitive. 

For each `log***` function in search page client, we also now need to expose a `build***` function that resurface the associated action cause, metadata, and the associated `log` function.

### DIFF
--- a/src/coveoua/library.ts
+++ b/src/coveoua/library.ts
@@ -4,10 +4,12 @@ import * as history from '../history';
 import * as SimpleAnalytics from './simpleanalytics';
 import * as storage from '../storage';
 export {CoveoAnalyticsClient, AnalyticsClientSendEventHook} from '../client/analytics';
-export {PreprocessAnalyticsRequest} from '../client/analyticsRequestClient';
+export {PreprocessAnalyticsRequest, AnalyticsClientOrigin} from '../client/analyticsRequestClient';
 export {IRuntimeEnvironment} from '../client/runtimeEnvironment';
 export {CoveoUA, getCurrentClient, handleOneAnalyticsEvent} from './simpleanalytics';
-export {CoveoSearchPageClient, SearchPageClientProvider} from '../searchPage/searchPageClient';
+export {CoveoSearchPageClient, SearchPageClientProvider, EventDescription} from '../searchPage/searchPageClient';
+export * from '../searchPage/searchPageEvents';
 export {CaseAssistClient} from '../caseAssist/caseAssistClient';
+export {SearchEventResponse} from '../events';
 
 export {analytics, donottrack, history, SimpleAnalytics, storage};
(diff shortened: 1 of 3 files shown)


src/searchPage/searchPageClient.ts
```diff
@@ -46,6 +46,13 @@ export interface SearchPageClientOptions extends ClientOptions {
     enableAnalytics: boolean;
 }
 
+export type EventDescription = Pick<SearchEventRequest, 'actionCause' | 'customData'>;
+
+export interface EventBuilder {
```
ThibodeauJF, line 51: If it's an event "builder" I would rename the "make" for "build" to stay consistent

btaillon (approved):

src/coveoua/library.ts
```diff
@@ -4,10 +4,17 @@ import * as history from '../history';
 import * as SimpleAnalytics from './simpleanalytics';
 import * as storage from '../storage';
 export {CoveoAnalyticsClient, AnalyticsClientSendEventHook} from '../client/analytics';
-export {PreprocessAnalyticsRequest} from '../client/analyticsRequestClient';
+export {PreprocessAnalyticsRequest, AnalyticsClientOrigin} from '../client/analyticsRequestClient';
 export {IRuntimeEnvironment} from '../client/runtimeEnvironment';
 export {CoveoUA, getCurrentClient, handleOneAnalyticsEvent} from './simpleanalytics';
-export {CoveoSearchPageClient, SearchPageClientProvider} from '../searchPage/searchPageClient';
+export {
```
olamothe, line 10: These were just not exported properly, so we'd have to do some un-needed gymnastic to get references to these types in other projects.

dlafreniere (approved):
is it necessary to have all of those builders directly in this library? Provided the required types are exported, could they be "helper functions" in headless?

olamothe:
> is it necessary to have all of those builders directly in this library? Provided the required types are exported, could they be "helper functions" in headless?

They could, but then Headless would have to do the mapping and transformation necessary to know that "when logging search box submit, the custom data contains X,Y,Z key with these values".

Which means we would necessarily need to copy paste some logic that already exists in `coveo.analytics` in order to build the correct format

And I would like to prevent having two source of truth for anything analytics related.

This was the case with JSUI + coveo.analytics combo in the past, where we've had quite a lot of problem for some commerce clients.

ThibodeauJF (approved):

btaillon (approved):

olamothe:
I will put this one on hold: The changes needed from Headless would introduce breaking changes.

So we will push back this one until this summer to coincide with next major version of Headless.

PR #327, opened 2022-02-21
[SFINT-4408] Metadata sent updated in logSelectDocumentSuggestion and logSelectFieldSuggestion

mmitiche:
### New properties needed  to be added to the metadata sent by logSelectDocumentSuggestion to report correctly the following events:
- Opening the Quickview of a document suggestion.
- Opening the document suggestion after clicking on its title from the Quickview.

### New property needed  to be added in the metadata sent by logSelectFieldSuggestion to report correctly the following event:

- Auto-selecting the classification with the highest confidence.

### DIFF
--- a/src/caseAssist/caseAssistActions.ts
+++ b/src/caseAssist/caseAssistActions.ts
@@ -71,6 +71,7 @@ export interface FieldSuggestion {
         value: string;
         confidence: number;
     };
+    autoSelection?: boolean;
 }
 
 export interface DocumentSuggestion {
@@ -83,4 +84,6 @@ export interface DocumentSuggestion {
         documentUrl: string;
         documentPosition: number;
     };
+    fromQuickview?: boolean;
+    openDocument?: boolean;
 }
--- a/src/caseAssist/caseAssistClient.spec.ts
+++ b/src/caseAssist/caseAssistClient.spec.ts
@@ -50,26 +50,33 @@ describe('CaseAssistClient', () => {
         },
     };
 
-    const fakeFieldSuggestion: FieldSuggestion = {
-        classification: {
-            value: 'some-field-value',
-            confidence: 0.5,
-        },
-        classificationId: 'field-suggestion-id',
-        fieldName: 'some-field',
-        responseId: 'field-suggestion-response-id',
+    const fakeFieldSuggestion = (autoSelection?: boolean): FieldSuggestion => {
+        return {
+            classification: {
+                value: 'some-field-value',
+                confidence: 0.5,
+            },
+            classificationId: 'field-suggestion-id',
+            fieldName: 'some-field',
+            responseId: 'field-suggestion-response-id',
+            ...(autoSelection && {autoSelection}),
+        };
     };
 
-    const fakeDocumentSuggestion: DocumentSuggestion = {
-        responseId: 'document-suggestion-response-id',
-        suggestionId: 'document-suggestion-id',
-        suggestion: {
-            documentPosition: 0,
-            documentTitle: 'the document title',
-            documentUri: 'some-document-uri',
-            documentUriHash: 'documenturihash',
-            documentUrl: 'some-document-url',
-        },
+    const fakeDocumentSuggestion = (fromQuickview?: boolean, openDocument?: boolean): DocumentSuggestion => {
+        return {
+            responseId: 'document-suggestion-response-id',
+            suggestionId: 'document-suggestion-id',
+            suggestion: {
+                documentPosition: 0,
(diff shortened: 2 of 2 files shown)


lbergeron (approved):
It looks good to me!

src/caseAssist/caseAssistActions.ts
```diff
@@ -83,4 +84,6 @@ export interface DocumentSuggestion {
         documentUrl: string;
         documentPosition: number;
     };
+    fromQuickview?: boolean;
+    openDocument?: boolean;
```
erocheleau, line 88: What's the difference between these two?

erocheleau: I think we should start documenting these interfaces? dunno if that's the standard, doesn't look like it is in this package...

src/caseAssist/caseAssistClient.spec.ts
```diff
@@ -50,26 +50,33 @@ describe('CaseAssistClient', () => {
...
+    const fakeFieldSuggestion = (autoSelection?: boolean): FieldSuggestion => {
+        return {
+            classification: {
+                value: 'some-field-value',
+                confidence: 0.5,
+            },
+            classificationId: 'field-suggestion-id',
+            fieldName: 'some-field',
+            responseId: 'field-suggestion-response-id',
+            ...(autoSelection && {autoSelection}),
```
erocheleau, line 62: I think I would prefer if the method set a default value and you were to change that value later in the test no?

```
const myFakeFieldSuggestion = fakeFieldSuggestion();
myFakeFieldSuggestion.autoSelection = true;
```

src/caseAssist/caseAssistClient.spec.ts
```diff
@@ -50,26 +50,33 @@ describe('CaseAssistClient', () => {
...
-        responseId: 'document-suggestion-response-id',
-        suggestionId: 'document-suggestion-id',
-        suggestion: {
-            documentPosition: 0,
-            documentTitle: 'the document title',
-            documentUri: 'some-document-uri',
-            documentUriHash: 'documenturihash',
-            documentUrl: 'some-document-url',
-        },
+    const fakeDocumentSuggestion = (fromQuickview?: boolean, openDocument?: boolean): DocumentSuggestion => {
```
erocheleau, line 66: Same as above, I would prefer that method to only set default and not take any parameter unless necessary.
Then you change the object received from that method.

erocheleau (changes requested):
Only test method that I would change and if you could explain what's the difference between the two parameters?

src/caseAssist/caseAssistActions.ts
mmitiche: fromQuickview property is set to true when the user opens the Quickview of a document suggestion and the openDocumentproperty is set to true when the user clicks on title of the document suggestion in the Quickview which  will trigger the opening of the document suggestion in a new tab.

src/caseAssist/caseAssistClient.spec.ts
mmitiche: Do you think this change will improve the clarity and the simplicity of the tests or there is another reason?

src/caseAssist/caseAssistActions.ts
erocheleau: I think we typically only differentiate between openQuickview or a regular click

1. Opening the quickview:
image

2. Clicking on the title of the quickview:
image

3. clicking on the document title the normal way
image

erocheleau: So TL;DR, clicking on the title of the quickview document and clicking on the document itself to open, are both the same "event". Only opening the quickview should be different IMO

erocheleau (changes requested):

src/caseAssist/caseAssistActions.ts
erocheleau: Misunderstanding, all is good :)

erocheleau (approved):

src/caseAssist/caseAssistActions.ts
mmitiche: That's a small summary of the events we wanna capture:
we have the QuanticDocumentSuggestion component which is an accordion containing different document suggestions.

- when expanding a document suggestion we send a suggestion_click event without any additional property.
- when clicking the quickview button of a document suggestion  we send a suggestion_click event with the property fromQuickview set to true.
- when clicking on the title of the document suggestion in the Quickview, this will open the document suggestion in a new tab, we send a suggestion_click event with the property openDocument set to true.

PR #332, opened 2022-03-21
KIT-1486 exported SmartSnippetFeedbackReason

btaillon:
https://coveord.atlassian.net/browse/KIT-1486

Source: https://github.com/coveo/ui-kit/pull/1843/files#r831436327

I'm not so sure about exporting this type from UA since I don't see a whole lot that's exported in general.

However, this is the only case I could find in the library of an enum being taken as a parameter for logging analytics. Unfortunately, enums aren't simply translated to an alias type, so that means that anyone using Typescript to call the library is expected to either provide the same enum (impossible since it's not exported) or work around that requirement by using utility types and a `as`.

### DIFF
--- a/src/coveoua/library.ts
+++ b/src/coveoua/library.ts
@@ -9,5 +9,6 @@ export {IRuntimeEnvironment} from '../client/runtimeEnvironment';
 export {CoveoUA, getCurrentClient, handleOneAnalyticsEvent} from './simpleanalytics';
 export {CoveoSearchPageClient, SearchPageClientProvider} from '../searchPage/searchPageClient';
 export {CaseAssistClient} from '../caseAssist/caseAssistClient';
+export {SmartSnippetFeedbackReason} from '../searchPage/searchPageEvents';
 
 export {analytics, donottrack, history, SimpleAnalytics, storage};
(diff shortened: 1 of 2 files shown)


src/coveoua/library.ts
```diff
@@ -9,5 +9,6 @@ export {IRuntimeEnvironment} from '../client/runtimeEnvironment';
 export {CoveoUA, getCurrentClient, handleOneAnalyticsEvent} from './simpleanalytics';
 export {CoveoSearchPageClient, SearchPageClientProvider} from '../searchPage/searchPageClient';
 export {CaseAssistClient} from '../caseAssist/caseAssistClient';
+export {SmartSnippetFeedbackReason} from '../searchPage/searchPageEvents';
```
olamothe, line 12: Just add it to the export 2 lines above.

It's already exporting types (SearchPageClientProvider).

olamothe (approved):

src/coveoua/library.ts
samisayegh: Since the enum was not exported in the past, could we adjust to a string union type instead?

samisayegh (approved):

PR #336, opened 2022-03-28
SFINT-4444 searchHub added to payload sent by the Case Assist analytics Client

mmitiche:
[SFINT-4444](https://coveord.atlassian.net/browse/SFINT-4444)

 - New optional parameter `provider` of type `CaseAssistClientProvider` added to instantiate the `CaseAssistClient`. This parameter will allow the client to add the search hub to the payload of all the case assist analytics events sent.

### DIFF
--- a/src/caseAssist/caseAssistClient.ts
+++ b/src/caseAssist/caseAssistClient.ts
@@ -16,6 +16,10 @@ import {
     UpdateCaseFieldMetadata,
 } from './caseAssistActions';
 
+export interface CaseAssistClientProvider {
+    getOriginLevel1: () => string;
+}
+
 export interface CaseAssistClientOptions extends ClientOptions {
     enableAnalytics?: boolean;
 }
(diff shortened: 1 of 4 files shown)


erocheleau (approved):

src/caseAssist/caseAssistClient.ts
```diff
@@ -108,10 +112,28 @@ export class CaseAssistClient {
...
 
     private sendClickEvent() {
-        return this.coveoAnalyticsClient.sendEvent('event', 'svc', CaseAssistEvents.click);
+        return this.coveoAnalyticsClient.sendEvent(
+            'event',
+            'svc',
+            CaseAssistEvents.click,
+            this.provider
+                ? {
+                      searchHub: this.provider.getOriginLevel1(),
```
nathanlb, line 134: If the value is given to key `searchHub` and in normal searches it's in `originLevel1` would filtering analytics to aggregate all matching searchHub values be a bit unintuitive?

mmitiche: Actually I used the key `searchHub` because this key is included in this list: [coveoExtensionsKeys](https://github.com/coveo/coveo.analytics.js/blob/81339db85f216f6439f2f02ff0cd5a4d86c7faf1/src/client/measurementProtocolMapping/baseMeasurementProtocolMapper.ts#L37-L47).  the list of keys that are supported by the Coveo collect protocol. `originiLevel1` is not in that list.

nathanlb: Well hot damn! Thanks for the info 👍

nathanlb (approved):

PR #341, opened 2022-05-03
Added full snippet as parameter for expand/collapse smart snippet suggestion

btaillon:
https://coveord.atlassian.net/browse/KIT-1590

Just logging the source is not enough, because there can be multiple related questions with the same source.


src/searchPage/searchPageEvents.ts
```diff
@@ -386,6 +386,12 @@ export interface QueryErrorMeta {
 export type SmartSnippetFeedbackReason = 'does_not_answer' | 'partially_answers' | 'was_not_a_question' | 'other';
 
 export interface SmartSnippetSuggestionMeta {
```
btaillon, line 388: not exported

ThibodeauJF (approved):
doesn't seem breaking 👍

olamothe (approved):

PR #344, opened 2022-05-06
Remove logShowMoreSmartSnippetSuggestion and logShowLessSmartSnippetSuggestion

btaillon:
https://coveord.atlassian.net/browse/KIT-1614

This is technically breaking if somebody is already using it, but it's so recent that it's very unlikely. If you're too worried about the breaking change, I can close this PR.

### DIFF
--- a/src/searchPage/searchPageClient.spec.ts
+++ b/src/searchPage/searchPageClient.spec.ts
@@ -563,6 +563,32 @@ describe('SearchPageClient', () => {
         });
     });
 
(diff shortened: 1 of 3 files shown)


ThibodeauJF:
Just check whether the breaking change has been released in prod. If we did release it, then just deprecate it

btaillon:
> Just check whether the breaking change has been released in prod. If we did release it, then just deprecate it

We did (at-least in NPM), I'll deprecate it then.

ThibodeauJF (approved):

olamothe (approved):

PR #352, opened 2022-06-20
feat: add instant results click event

liviarett:
https://coveord.atlassian.net/browse/KIT-1775

[EDIT]:

after our conversation during stand up today, I'm thinking of sending an extra bit of `customData` called `location` (happy to change name and value upon request) so that the event is the same as a regular result click, but you can differentiate by location

Screenshot 2022-06-21 at 17 07 26

the default would be not to have a location, or we can add a `mainResultList` via atomic

cc @lvu285

### DIFF
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -25,6 +25,7 @@ import {
     SmartSnippetDocumentIdentifier,
     StaticFilterMetadata,
     StaticFilterToggleValueMetadata,
+    DocumentMetadata,
 } from './searchPageEvents';
 import {NoopAnalytics} from '../client/noopAnalytics';
 import {formatOmniboxMetadata} from '../formatting/format-omnibox-metadata';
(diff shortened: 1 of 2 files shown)


lvu285:
evenType is perfect for me.
Btw, wonder if we will add the UA for "Show all results" button in Preview as well?

src/searchPage/searchPageClient.ts
```diff
@@ -328,6 +328,13 @@ export class CoveoSearchPageClient {
         return this.logCustomEvent(SearchPageEvents.clearRecentQueries);
     }
 
+    public logInstantResultClick(info: PartialDocumentInformation, identifier: DocumentIdentifier) {
+        return this.logCustomEvent(SearchPageEvents.instantResultClick, {
```
olamothe, line 332: Needs to be a click, not a custom event.

olamothe:
I would contest the original bug: https://coveord.atlassian.net/browse/KIT-1767

A normal click event (documentOpen) is perfectly fine.

src/searchPage/searchPageEvents.ts
```diff
@@ -296,6 +300,7 @@ export const CustomEventsTypes: Partial<Record<SearchPageEvents, string>> = {
     [SearchPageEvents.recentResultClick]: 'recentlyClickedDocuments',
     [SearchPageEvents.clearRecentResults]: 'recentlyClickedDocuments',
     [SearchPageEvents.showLessFoldedResults]: 'folding',
+    [SearchPageEvents.instantResultClick]: 'instantResults',
```
olamothe, line 303: Changing to a click event means this is not needed anymore.

PR #359, opened 2022-07-13
feat(ip): logContextChanged & logExpandToFullUI events added to the insight client

mmitiche:
## 1. logContextChanged

In the service console inside Salesforce, when a user updates one of the case fields, the insight panel should automatically trigger a new search query and send the ContextChanged search event.

A new `logContextChanged` method is now added to the insight client.

- #### Current ContextChanged event in the C4SF package:

PreviousContextChangedEvent

- #### New ContextChanged event sent using coveo.analytics.js:

MewcontextEvent2

## 2. logExpandToFullUI

In the service console inside Salesforce, when a user opens the full search page, the Hosted Insight Panel should send the ExpandToFullUI  custom event.

A new `logExpandToFullUI` method is now added to the insight client.

- #### Current ExpandToFullUI event in the C4SF package:

CurrentExpandToFullUIEvent

- #### New ExpandToFullUI event sent using coveo.analytics.js:

NewOpenUI2

### DIFF
--- a/src/insight/insightClient.spec.ts
+++ b/src/insight/insightClient.spec.ts
@@ -256,7 +256,7 @@ describe('InsightClient', () => {
         const meta = {
             caseId: '1234',
             caseNumber: '5678',
-            caseContext: {subject: 'test subject', description: 'test description'},
+            caseContext: {Case_Subject: 'test subject', Case_Description: 'test description'},
         };
 
         const metaToBeSent = {
@@ -275,7 +275,7 @@ describe('InsightClient', () => {
         const meta = {
             caseId: '1234',
             caseNumber: '5678',
-            caseContext: {subject: 'test subject', description: 'test description'},
+            caseContext: {Case_Subject: 'test subject', Case_Description: 'test description'},
             fullSearchComponentName: 'c__FullSearch',
             triggeredBy: 'openFullSearchButton',
         };
--- a/src/insight/insightClient.spec.ts
+++ b/src/insight/insightClient.spec.ts
@@ -259,7 +259,7 @@ describe('InsightClient', () => {
             caseContext: {Case_Subject: 'test subject', Case_Description: 'test description'},
         };
 
-        const metaToBeSent = {
+        const expectedMeta = {
             CaseId: '1234',
             CaseNumber: '5678',
             CaseSubject: 'test subject',
@@ -268,7 +268,7 @@ describe('InsightClient', () => {
         };
 
         await client.logContextChanged(meta);
-        expectMatchPayload(InsightEvents.contextChanged, metaToBeSent);
+        expectMatchPayload(InsightEvents.contextChanged, expectedMeta);
     });
 
     it('should send proper payload for #expandToFullUI', async () => {
@@ -280,7 +280,7 @@ describe('InsightClient', () => {
             triggeredBy: 'openFullSearchButton',
         };
 
-        const metaToBeSent = {
+        const expectedMeta = {
             CaseId: '1234',
             CaseNumber: '5678',
             CaseSubject: 'test subject',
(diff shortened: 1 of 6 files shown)


aelfaleh (approved):

erocheleau (approved):

erocheleau:
Are you merging this PR with this one as well?
https://coveord.atlassian.net/browse/SFINT-4570

mmitiche:
> Are you merging this PR with this one as well?
> https://coveord.atlassian.net/browse/SFINT-4570

Yes I'm planning to do that.

src/insight/insightEvents.ts
```diff
@@ -0,0 +1,23 @@
+export enum InsightEvents {
+    /**
+     * Identifies the search event that gets logged when the query context is updated as a result of updating one of the case context fields.
+     */
+    contextChanged = 'contextChanged',
+    /**
+     * Identifies the event that gets logged when the user opens the full search page from the insight panell.
```
mmitiche, line 7:
```suggestion
     * Identifies the event that gets logged when the user opens the full search page from the insight panel.
```

mmitiche:
> It all looks pretty good to me. Using different format for the `customData > caseContext` information might be a problem for reporting though. Including the data of the hosted insight panel into an existing report would be problematic.
> 
> @erocheleau Any thoughts on this?

Totally right! I'm currently working on it to assure that we keeping the exact same format.

mmitiche:
> > It all looks pretty good to me. Using different format for the `customData > caseContext` information might be a problem for reporting though. Including the data of the hosted insight panel into an existing report would be problematic.
> > @erocheleau Any thoughts on this?
> 
> Totally right! I'm currently working on it to assure that we keeping the exact same format.

The events are now updated to exactly match what we currently send in the C4SF package.

src/insight/insightClient.spec.ts
```diff
@@ -253,20 +253,41 @@ describe('InsightClient', () => {
...
 
     it('should send proper payload for #contextChanged', async () => {
-        const meta = {caseId: '1234', caseNumber: '1234', caseContext: {subject: 'test'}};
+        const meta = {
+            caseId: '1234',
+            caseNumber: '5678',
+            caseContext: {subject: 'test subject', description: 'test description'},
+        };
+
+        const metaToBeSent = {
```
lbergeron, line 262: I'd suggest to name the variable `expectedMeta`. It seems more meaningful to me.

src/insight/insightClient.ts
```diff
@@ -34,6 +34,17 @@ export interface InsightClientOptions extends ClientOptions {
     enableAnalytics: boolean;
 }
 
+const extractContextFromMetadata = (meta: {caseContext: Record<string, string>}) => {
+    const context: Record<string, string> = {};
+    Object.keys(meta.caseContext).forEach((contextKey) => {
+        const capitalizedKey = contextKey.charAt(0).toUpperCase() + contextKey.slice(1);
+        const keyToBeSent = `context_Case_${capitalizedKey}`;
```
lbergeron, line 41: I'm not sure about adding this `Case_` prefix on all context values. I could imagine an implementation adding some context information that is not directly related to the case. It might look weird to have `Case` in the context key name.

Maybe we could add the `Case_` prefix only for the keys that are well known like `subject` and `description`.

mmitiche: I removed `Case_` prefix. Now we just send `context_${contextKey}`.
If we want to have a `context_Case_Subject` key in the event payload we only have to send a context containing a key named `Case_Subject`

lbergeron: So it'll be the Hosted Insight Panel component responsability to use the same key as before to ensure a consistent reporting.

mmitiche: Yes exactly 👌I will add tests at the Hosted Insight Panel component level to confirm that we always set the correct context keys that will ensure a consistent reporting.

PR #365, opened 2022-08-02
KIT-1902 Update react native warning

ThibodeauJF:
Without breaking, this is how the ReactNative runtime can be used

- the path was wrong, the react-native subfolder is inside the src, [which is exported ](https://github.com/coveo/coveo.analytics.js/blob/master/package.json#L67)
- the client is CoveoAnalyticsClient, not CoveoAnalytics
- there is a required param for the constructor of ReactNativeRuntime

We should definitely improve this in a major version

### DIFF
--- a/src/react-native/react-native-runtime.ts
+++ b/src/react-native/react-native-runtime.ts
@@ -9,6 +9,7 @@ export class ReactNativeRuntime implements IRuntimeEnvironment {
     public client: AnalyticsFetchClient;
 
     // TODO: v3 switch to ClientOptions type, add default options
+    // TODO: v3 reuse own ReactNativeStorage to implement VisitorIdProvider's getCurrentVisitorId, setCurrentVisitorId
     constructor(clientOptions: IAnalyticsClientOptions) {
         this.storage = new ReactNativeStorage();
         this.client = new AnalyticsFetchClient(clientOptions);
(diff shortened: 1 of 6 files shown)


btaillon (approved):

src/react-native/react-native-utils.ts
```diff
@@ -3,11 +3,13 @@ export const ReactNativeRuntimeWarning = `
...
-        import {ReactNativeRuntime} from 'coveo.analytics/react-native';
+        import {ReactNativeRuntime} from 'coveo.analytics/src/react-native';
         
-        const analytics = new CoveoAnalytics({
+        const analytics = new CoveoAnalyticsClient({
             ...your options,
-            runtimeEnvironment: new ReactNativeRuntime();
+            runtimeEnvironment: new ReactNativeRuntime({
+                baseUrl: '...',
+            });
```
samisayegh, line 12: Is there a test we could write to make sure the package is working as expecting with react-native? It feels a test could have caught the incorrect import path (if not the typescript build process)?

In case it's helpful, there is documentation here about mocking the [react-native async storage module](https://react-native-async-storage.github.io/async-storage/docs/advanced/jest/). Maybe it could be useful, not sure 🤞

src/react-native/react-native-runtime.ts
```diff
@@ -4,6 +4,7 @@ import {ReactNativeStorage} from './react-native-storage';
 import {IRuntimeEnvironment} from '../client/runtimeEnvironment';
 import {IAnalyticsClientOptions} from '../client/analyticsRequestClient';
 
+// TODO: improve usability & document use in major version bump
```
samisayegh, line 7: This comment feels vague for a future contributor to act on. Is there something specific we could highlight? If not, it may be best to remove

samisayegh (approved):

PR #366, opened 2022-08-03
Make "view" events send clientId inside payload

ThibodeauJF:
Headless (SearchPageClient) properly sends the client id as part of the payload, but not the simple analytics client & its view event which we make customer use to register view events

- https://docs.coveo.com/en/headless/latest/usage/headless-usage-analytics/headless-view-events/
- https://coveord.atlassian.net/browse/KIT-1846
- https://discuss.coveo.com/t/support-00079621-clientid-null-in-view-event-page-views-not-tracked-properly-as-a-result/7922/22
- https://discuss.coveo.com/t/migration-from-visitor-id-to-client-id/8217

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -54,6 +54,7 @@ export type EventTypeConfig = {
     newEventType: EventType;
     variableLengthArgumentsNames?: string[];
     addVisitorIdParameter?: boolean;
+    addClientIdParameter?: boolean;
     usesMeasurementProtocol?: boolean;
 };
 
@@ -75,6 +76,7 @@ export interface AnalyticsClient {
      * @deprecated
      */
     readonly currentVisitorId: string;
+    getCurrentVisitorId?(): Promise<string>; // TODO: v3 make required
 }
 
 export interface BufferedRequest {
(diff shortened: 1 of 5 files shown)


src/client/analytics.ts
```diff
@@ -205,8 +207,12 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
     }
 
     async resolveParameters(eventType: EventType | string, ...payload: VariableArgumentsPayload) {
-        const {variableLengthArgumentsNames = [], addVisitorIdParameter = false, usesMeasurementProtocol = false} =
-            this.eventTypeMapping[eventType] || {};
+        const {
+            variableLengthArgumentsNames = [],
+            addVisitorIdParameter = false,
+            usesMeasurementProtocol = false,
+            addClientIdParameter = false,
```
samisayegh, line 214: Could the default be `true` here?

src/coveoua/simpleanalytics.spec.ts
```diff
@@ -199,7 +208,16 @@ describe('simpleanalytics', () => {
 
             expect(fetchMock.calls().length).toBe(1);
             expect(fetchMock.lastUrl()).toBe(`${analyticsEndpoint}/pageview`);
-            expect(JSON.parse(fetchMock.lastCall()[1].body.toString())).toEqual({somedata: 'asd'});
+            expect(JSON.parse(fetchMock.lastCall()![1]!.body!.toString())).toEqual({somedata: 'asd'});
+        });
+
+        it.only('can send view event with clientId', async () => {
```
samisayegh, line 214: Let's remove the `only` call here,

src/client/analytics.ts
```diff
@@ -75,6 +76,7 @@ export interface AnalyticsClient {
      * @deprecated
      */
     readonly currentVisitorId: string;
+    getCurrentVisitorId?(): Promise<string>; // TODO: v3 make required
```
samisayegh, line 79: I don't think the `AnalyticsClient` interface exported publicly to consumers of the package. If so, maybe the method can be required?

samisayegh (approved):

src/coveoua/simpleanalytics.spec.ts
ThibodeauJF: damn!

src/client/analytics.ts
ThibodeauJF: I'd rather not create undesired side effects. I've notice the client Id is already correctly present in the payload for SearchPageClient

src/client/analytics.ts
ThibodeauJF: For some reason everything under src is part of the published NPM package, this means technically users can implement the interface.

btaillon (approved):

olamothe (approved):

PR #368, opened 2022-08-15
KIT-1932 convert logUndoTriggerQuery to search event

btaillon:
https://coveord.atlassian.net/browse/KIT-1932

Pressing "undo" triggers a query.


ThibodeauJF (approved):

btaillon:
Adding @jpmarceau in case he has opinions on using a custom event vs search box event for this.

jpmarceau (approved):
I think this makes sense.
In the doc, we generally write that anything that triggers a query is a search event, so this qualifies 🙂

PR #377, opened 2022-10-28
SFINT-4661 logDocumentOpen added to the insight analytics client

mmitiche:
[SFINT-4661](https://coveord.atlassian.net/browse/SFINT-4661)

### Example:
HeadlessClickEvent

### DIFF
--- a/src/insight/insightClient.ts
+++ b/src/insight/insightClient.ts
@@ -221,7 +221,7 @@ export class CoveoInsightClient {
 
     public logDocumentOpen(info: PartialDocumentInformation, identifier: DocumentIdentifier, metadata?: CaseMetadata) {
         if (metadata) {
-            const metadataToSend = generateMetadataToSend(metadata);
+            const metadataToSend = generateMetadataToSend(metadata, false);
             return this.logClickEvent(SearchPageEvents.documentOpen, info, identifier, metadataToSend);
         }
         return this.logClickEvent(SearchPageEvents.documentOpen, info, identifier);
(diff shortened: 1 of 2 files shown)


nathanlb (approved):
Looks good to me. Can we think of any other event we'll need for the GA features of the hosted insight panel?

mmitiche:
> Looks good to me. Can we think of any other event we'll need for the GA features of the hosted insight panel?

I think we are covering pretty much everything, but I will do a second pass on all the search analytics events to make sure we are not missing any that should also be present for the insight use case.

src/insight/insightClient.ts
```diff
@@ -213,6 +219,14 @@ export class CoveoInsightClient {
         return this.logCustomEvent(InsightEvents.expandToFullUI, metadataToSend);
     }
 
+    public logDocumentOpen(info: PartialDocumentInformation, identifier: DocumentIdentifier, metadata?: CaseMetadata) {
+        if (metadata) {
+            const metadataToSend = generateMetadataToSend(metadata, false);
+            return this.logClickEvent(SearchPageEvents.documentOpen, info, identifier, metadataToSend);
+        }
+        return this.logClickEvent(SearchPageEvents.documentOpen, info, identifier);
+    }
```
lbergeron, line 228: We can avoid repeating the `logClickEvent` arguments by using a ternary. IMO it is a little more readable.

```suggestion
        return this.logClickEvent(
          SearchPageEvents.documentOpen,
          info,
          identifier,
          metadata ? generateMetadataToSend(metadata, false) : undefined);
    }
```

lbergeron (approved):

PR #379, opened 2022-10-31
Improve React Native support

ThibodeauJF:
**[Usage example in app](https://github.com/ThibodeauJF/coveo-analytics-react-native/blob/master/App.js)** 

This is slightly breaking, but I'm only breaking it because the previous implementation simply didn't work, to import CoveoAnalytics inside react native you would need to do the following:

```
import {CoveoAnalyticsClient} from 'coveo.analytics/dist/library.es';
```

With this PR there is now a new sub package where you can import everything you need:
```
import {
  ReactNativeRuntime,
  CoveoAnalyticsClient,
  ...
} from 'coveo.analytics/react-native';
```

I also rearranged the options so they make more sense, before you would need to implement your own "visitorIdProvider", build your own "baseUrl" from the endpoint & versions. It's not something that's straightforward or dev friendly.

**I removed the hard dependency "@react-native-async-storage/async-storage" for multiple reason:**
- It's a package 99.9% of the lib's users won't need.
- It came with a lot of vulnerabilities.
- It's support inside an OOTB React Native project is terrible, you get an error because of it OOTB (see image) that requires additional work to be fixed (Especially with expo where you need to "eject" the app to use it properly).
- There are alternative libraries, we shouldn't force the use of this one.
- Implementing a custom Storage is pretty easy.

storage


src/client/analytics.ts
```diff
@@ -469,9 +474,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
     }
 
     private get baseUrl(): string {
-        const {version, endpoint} = this.options;
-        const endpointIsCoveoProxy = endpoint.indexOf('.cloud.coveo.com') !== -1;
-        return `${endpoint}${endpointIsCoveoProxy ? '' : '/rest'}/${version}`;
+        return buildBaseUrl(this.options.version, this.options.version);
```
btaillon, line 477:
```suggestion
        return buildBaseUrl(this.options.endpoint, this.options.version);
```

btaillon (approved):
I like this approach better, but I'm a bit concerned about the breaking changes 🤔 

You said that in order to use the library prior to these changes, a weird import was required, but I don't think that guarantees nobody is doing it. This should probably be a minor version bump instead of a patch bump.

ThibodeauJF:
>  This should probably be a minor version bump instead of a patch bump.

It is "version": "2.22.0",

src/react-native/react-native-runtime.ts
```diff
@@ -1,17 +1,47 @@
...
+import {uuidv4} from '../client/crypto';
+import {buildBaseUrl} from '../client/analytics';
+
+export type ReactNativeStorage = WebStorage;
+
+export interface ReactNativeRuntimeOptions {
+    /**
+     * Mandatory Storage implementation.
+     *
+     * It is recommended to use AsyncStorage from the "@react-native-async-storage/async-storage"
```
jpmarceau, line 14:
```suggestion
     * We recommend using AsyncStorage from the "@react-native-async-storage/async-storage"
```

src/react-native/react-native-runtime.ts
```diff
@@ -1,17 +1,47 @@
...
+     * community package or the deprecated AsyncStorage from "react-native".
```
jpmarceau, line 15: Curious: why do we recommend using a deprecated package? If we recommend it, I'm worried we'll have to support it.

README.md
```diff
@@ -138,15 +138,58 @@ npm install coveo.analytics isomorphic-fetch
...
     eventType: 'dog',
     eventValue: 'Hello! Yes! This is Dog!',
+    language: 'en',
+});
+```
+
+## Using React Native
+
+Since React Native is not running inside a browser, it cannot use cookies or the local/session storage that modern browsers provide.
+It is necessary to provide your own Storage implementation. Thankfully there exist multiple packages to store data:
```
jpmarceau, line 156:
```suggestion
Since React Native does not run inside a browser, it cannot use cookies or the local/session storage that modern browsers provide.
You must provide your own Storage implementation. Thankfully, there exist multiple packages to store data:
```

olamothe (approved):

src/react-native/react-native-runtime.ts
ThibodeauJF: It's a possible option depending on the customers setup & limitations (e.g. react native version they use). We won't have to support it or any of the package, since we just tell the customer to implement the interface (using any technology).

jpmarceau: Yeah I this it would be better to only recommend one thing, namely `react-native-async-storage` 👍

jpmarceau (approved):

PR #381, opened 2022-11-03
KIT-2055 Made trigger notify and execute events plural

btaillon:
https://coveord.atlassian.net/browse/KIT-2055

Technically breaking, but I checked with ML and UA, and neither are using those events. Plus, if somebody is calling this function and updates `coveo.analytics`, it's a very simple/obvious fix.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.22.3",
+    "version": "2.23.0",
     "description": "📈 Coveo analytics client (node and browser compatible) ",
     "main": "dist/library.js",
     "module": "dist/library.es.js",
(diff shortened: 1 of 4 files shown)


olamothe (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "coveo.analytics",
-    "version": "2.22.3",
```
olamothe, line 3: ? I dont think you need to change that manually.

btaillon: it's because I want it to be a minor upgrade instead of a patch

ThibodeauJF: package lock should be update too

ThibodeauJF (approved):

PR #384, opened 2022-11-18
fix: handle react-native no-document

jpdery26:
BasePlugin.ts fails in ReactNative (no document object)


btaillon (approved):
Looks good, without a document I agree that `lastReferrer` and `title` don't make much sense.

ThibodeauJF (approved):

ThibodeauJF:
Document might be the first but it might not be the only one that fails... window is another one that comes to mind

PR #386, opened 2022-11-25
Fixing currency symbol not being stripped when sending price [COM-1824]

toussaints:
Following [this discuss post](https://discuss.coveo.com/t/product-price-symbol-is-not-removed/8800) we saw that the code was not following the behaviour we described in the documentation:

>Any text surrounding the value, such as a currency symbol, will be removed before it’s sent to [Coveo Usage Analytics (Coveo UA)](https://docs.coveo.com/en/182/). For example, when €-20.00 is entered, only -20.00 is sent.

This PR aims at correcting that.

**Added the test first:**
image

🎩
**After the fix:**
image

Also ran the lintfix.

### DIFF
--- a/src/plugins/ec.ts
+++ b/src/plugins/ec.ts
@@ -34,7 +34,7 @@ export interface ProductProperties extends CoveoExtensionProperties {
     brand?: string;
     category?: string;
     variant?: string;
-    price?: number;
+    price?: Number | string;
     quantity?: number;
     coupon?: string;
     position?: number;
@@ -51,7 +51,7 @@ export interface ImpressionProperties extends CoveoExtensionProperties {
     category?: string;
     variant?: string;
     position?: number;
-    price?: number;
+    price?: any;
     custom?: CustomValues;
 }
 
@@ -144,6 +144,7 @@ export class ECPlugin extends BasePlugin {
     private getProductPayload() {
         return this.products
             .map((product) => this.assureProductValidity(product))
+            .map((product) => this.convertProductPriceToNumber(product))
             .reduce((newPayload, product, index) => {
                 return {
                     ...newPayload,
@@ -159,9 +160,9 @@ export class ECPlugin extends BasePlugin {
                 ({impressions, ...rest}) =>
                     ({
                         ...rest,
-                        impressions: impressions.map((baseImpression) =>
-                            this.assureBaseImpressionValidity(baseImpression)
-                        ),
+                        impressions: impressions
+                            .map((baseImpression) => this.assureBaseImpressionValidity(baseImpression))
+                            .map((impression) => this.convertImpressionPriceToNumber(impression)),
                     } as ImpressionList)
             )
             .reduce((newPayload, impressionList, index) => {
@@ -170,6 +171,33 @@ export class ECPlugin extends BasePlugin {
                     ...convertImpressionListToMeasurementProtocol(impressionList, index, 'pi'),
                 };
             }, {});
+    }
+
+    private convertProductPriceToNumber(product: Product) {
+        if (product.price) {
+            product.price = this.tryConvertStringPriceToNumber(product.price);
+        }
+
+        return product;
+    }
+
+    private convertImpressionPriceToNumber(impression: BaseImpression) {
+        if (impression.price) {
+            impression.price = this.tryConvertStringPriceToNumber(impression.price);
+        }
+
+        return impression;
+    }
+
+    private tryConvertStringPriceToNumber(price: Number | string): Number | string {
+        if (typeof price === 'string') {
+            let parsedPrice = Number(price.replace(/[^0-9\.-]/g, ''));
+            if (parsedPrice) {
+                return parsedPrice;
+            }
+        }
+
+        return price;
     }
 
     private assureProductValidity(product: Product) {
--- a/src/plugins/ec.ts
+++ b/src/plugins/ec.ts
@@ -34,7 +34,7 @@ export interface ProductProperties extends CoveoExtensionProperties {
     brand?: string;
     category?: string;
     variant?: string;
-    price?: Number | string;
+    price?: number | string;
     quantity?: number;
     coupon?: string;
     position?: number;
@@ -51,7 +51,7 @@ export interface ImpressionProperties extends CoveoExtensionProperties {
     category?: string;
     variant?: string;
     position?: number;
-    price?: any;
+    price?: number | string;
     custom?: CustomValues;
 }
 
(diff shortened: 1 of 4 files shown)


src/plugins/ec.ts
```diff
@@ -172,6 +173,33 @@ export class ECPlugin extends BasePlugin {
...
+
+    private convertImpressionPriceToNumber(impression: BaseImpression) {
+        if (impression.price) {
+            impression.price = this.tryConvertStringPriceToNumber(impression.price);
+        }
+
+        return impression;
+    }
+
+    private tryConvertStringPriceToNumber(price: Number | string): Number | string {
```
Spuffynism, line 192: I think we can leave these as `number | string` because when we create a Number, we don't use the `new` keyword. Hence, it creates a primitive number which is compatible with `number`
image

```suggestion
    private tryConvertStringPriceToNumber(price: number | string): number | string {
```

This means we don't need to change `ProductProperties`'s `price` to a `Number | string`, but can put it as `number | string`. It should be more consistent with how the other properties that are `number`s are defined.

src/plugins/ec.ts
```diff
@@ -172,6 +173,33 @@ export class ECPlugin extends BasePlugin {
...
+        if (typeof price === 'string') {
+            let parsedPrice = Number(price.replace(/[^0-9\.-]/g, ''));
+            if (parsedPrice) {
+                return parsedPrice;
+            }
+        }
+
+        return price;
```
Spuffynism, line 200: nit: we could remove the nesting a bit by flipping the condition into a guard clause
```suggestion
        if (typeof price === 'number') {
            return price
        }
        
        return Number(price.replace(/[^0-9\.-]/g, '')) || price;
```

src/plugins/ec.ts
```diff
@@ -172,6 +173,33 @@ export class ECPlugin extends BasePlugin {
...
+        if (impression.price) {
+            impression.price = this.tryConvertStringPriceToNumber(impression.price);
+        }
+
+        return impression;
+    }
+
+    private tryConvertStringPriceToNumber(price: Number | string): Number | string {
+        if (typeof price === 'string') {
+            let parsedPrice = Number(price.replace(/[^0-9\.-]/g, ''));
```
Spuffynism, line 194: Are commas tolerated in prices? 
Should a price specified as `20 000` be transformed to `20000` or should it stay as its original form?

Should we instead follow the logic described by the [measurement protocol doc](https://developers.google.com/analytics/devguides/collection/protocol/v1/reference#currency)?
> Once the value is sent to Google Analytics, all text is removed up until the first digit, the `-` character or the `.` (decimal) character.

Spuffynism (approved):
Does this break current implementations that expect the dollar sign to not be stripped out?

src/plugins/ec.ts
thudon-coveo: I agree with mister Nick

src/plugins/ec.ts
```diff
@@ -51,7 +51,7 @@ export interface ImpressionProperties extends CoveoExtensionProperties {
     category?: string;
     variant?: string;
     position?: number;
-    price?: number;
+    price?: any;
```
thudon-coveo, line 54: Why are we using `any` here  instead of` number | string ` like the other price above ?

toussaints: Overlook, I'll fix it

src/plugins/ec.ts
```diff
@@ -172,6 +173,30 @@ export class ECPlugin extends BasePlugin {
...
+
+        return impression;
+    }
+
+    private tryConvertStringPriceToNumber(price: number | string): number | string {
+        if (typeof price === 'number') {
+            return price;
+        }
+
+        return Number(price.replace(/[^0-9\.-]/g, '')) || price;
```
Spuffynism, line 197: I think we unfortunately have to handle a number being zero. Otherwise, if the number is zero, it'll fallback to `price` 🤔. Sorry, should've caught that sooner.
```suggestion
        const parsedNumber = Number(price.replace(/[^0-9\.-]/g, ''));
        return parsedNumber != 0 ? parsedNumber : price;
```

toussaints: I'll add a test case

toussaints:
This would be considered the final version ready to merge

src/plugins/ec.ts
```diff
@@ -172,6 +173,32 @@ export class ECPlugin extends BasePlugin {
...
+
+        return impression;
+    }
+
+    private tryConvertStringPriceToNumber(price: number | string): number | string {
+        if (typeof price === 'number') {
+            return price;
+        }
+
+        let parsedPrice = parseFloat(price.replace(/[^0-9\.-]/g, ''));
```
Spuffynism, line 197: This can be a `const` since it isn't modified after its declaration
```suggestion
        const parsedPrice = parseFloat(price.replace(/[^0-9\.-]/g, ''));
```

thudon-coveo (approved):
Perfection

fvanham-coveo:
Thanks for putting this out there. I have a number of concerns, 2 of which are high level, one of which speaks to the regex itself:

1. In general, the data platform already has server side validation on currency payloads. Meaning we have substantially the same regex on the server end already (except that we don’t strip postfixes, to align the with GMP definition). Postfixed non-numbers will be NULLed because they are not up to spec, you’ll see (in due time) a syntax warning in the visit browser. With this PR we'd have two different validation rules in place in two different libraries, that don't 100% align. It's better to centralize these.
2. I’m ok with sending a in browser warning if a number possibly doesn’t parse, but moving payload modification to the client side is risky: the SSoT table will no longer be capturing what the customer sent us, but rather what a client side library in between sent to us. This becomes very opaque at one point, especially given the different implementation paths: e.g. if I use coveoua and send a payload, I get a different result logged than if I use straight-api and send the same payload. I’m worried this is going to create confusion down the line.
3. There’s weird side effects from this regexp that are now coming our way, that we have no way to detect, and we don’t have a log of the “original” values, which may have been correct but just incorrectly formatted. So the platform cannot inform the customer they sent in an incorrect number, AND we "corrected" :) it to an entirely different value, which will throw off all computations. Here's a sampling
- `2.000,30` (euro notation with thousand separator) becomes `2.00030` and customer angry because sales decreased
- `5.78E01` (scientific notation) becomes `5.7801`
- `5.67E-01` (scientific notation) becomes `5.67-01`
- `5,23453` (euro notation) becomes `523453` customer will be happy their sales increased ;)

@jmprovencher for visibility. Instead of fixing it here, I recommend we fix the docs. Payload fixing should happen AFTER we logged the client payload.

toussaints:
Agreed with @fvanham-coveo's comment to fix the documentation instead, thanks

PR #389, opened 2022-12-12
KIT-2071 made preparable analytics prepare everything

btaillon:
https://coveord.atlassian.net/browse/KIT-2071

# The old problem
In Atomic, we wish to include the `actionCause` and `customData` of analytics events in their corresponding search (`/v2`) call. This problem was solved in an old PR by making events "preparable". Events were made "preparable" by making a copy of each `log*` function in `coveo.analytics`'s `searchPageClient.ts`, respectively named `make*`, which returns the needed metadata and a function to log them.

# The new problem
In Atomic, analytics middleware (which are passed as `beforeSendHooks` in `coveo.analytics`) weren't applied to prepared events. This meant that the `customData` we sent in search calls could be different from the one in analytics calls. One way this manifested was that `coveoAtomicVersion` was missing in search calls.

# The solution
Since `beforeSendHooks` live at a lower level than `searchPageClient.ts` (they live in `analytics.ts`), and since there's a lot of other things that seem to modify events at that level, I decided to make events preparable in `analytics.ts`.

This had two side-effects:
1. Preparing analytics is now asynchronous, since various processes affecting the payload in `analytics.ts` are asynchronous.
2. The entire analytics payload is now "prepared", not just the "customData".

# The second new problem
With the above changes, since the entire analytics payload is now prepared, the `searchQueryUid` was also prepared. Unfortunately, that's not something we can do, since we intend to prepare analytics before their corresponding search call.

# The solution to the second new problem
I allowed prepared events in `analytics.ts` to take two different payloads:
* the prepared payload (e.g., `PreparedClickEventRequest`)
* the full payload (e.g., `ClickEventRequest`)

Only the prepared payload is required to prepare an event. To call a prepared event, the remaining payload is required.

In addition to this, I made all `make*` functions in `searchPageClient.ts` require the `searchUID` when logging them.

# Conclusion / tl;dr
Prior to this PR, the signature of `make` functions (which only exist in `searchPageClient.ts`, so no case assist nor insight events) looked like the following:
```typescript
export interface EventBuilder<T extends AnyEventResponse = AnyEventResponse> {
    description: {
      actionCause: string;
      customData?: Record<string, any>;
    };
    log: () => Promise<T | void>;
}

makeCollapseSmartSnippetSuggestion(...someParams): EventBuilder;
```

After this PR, the signature of `make` functions in the same file looks like the following:
```typescript
export interface EventBuilder<T extends AnyEventResponse = AnyEventResponse> {
    description: {
      actionCause: string;
      customData?: Record<string, any>;
    };
    log: (searchUID: string) => Promise<T | void>; // Now takes the `searchUID` as a parameter.
}

makeCollapseSmartSnippetSuggestion(...someParams): Promise<EventBuilder>; // Now returns a `Promise`.
```

This change is breaking, but in theory we don't really expect any of the affected functions to have been used anywhere but in Headless v2.

### DIFF
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -742,7 +742,7 @@ export class CoveoSearchPageClient {
             eventType,
             eventValue: event,
         };
-        const preparedEvent = await this.coveoAnalyticsClient.prepareCustomEvent(request);
+        const preparedEvent = await this.coveoAnalyticsClient.makeCustomEvent(request);
         return {
             description: this.makeEventDescription(preparedEvent, event),
             log: (searchUID) => preparedEvent.log({lastSearchQueryUid: searchUID}),
@@ -768,7 +768,7 @@ export class CoveoSearchPageClient {
             eventType,
             eventValue,
         };
-        const preparedEvent = await this.coveoAnalyticsClient.prepareCustomEvent(payload);
+        const preparedEvent = await this.coveoAnalyticsClient.makeCustomEvent(payload);
         return {
             description: this.makeEventDescription(preparedEvent, eventValue as SearchPageEvents),
             log: (searchUID) => preparedEvent.log({lastSearchQueryUid: searchUID}),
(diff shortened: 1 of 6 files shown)


src/searchPage/searchPageClient.ts
```diff
@@ -64,7 +65,7 @@ export type EventDescription = Pick<SearchEventRequest, 'actionCause' | 'customD
 
 export interface EventBuilder<T extends AnyEventResponse = AnyEventResponse> {
     description: EventDescription;
-    log: () => Promise<T | void>;
+    log: (searchUID: string) => Promise<T | void>;
```
ThibodeauJF, line 68: I'd pass an object here so it could be updated without changing adding an argument

ThibodeauJF (approved):

olamothe (approved):

PR #390, opened 2022-12-16
[UA-7204] Move coveoua custom object from custom to customData for non-collect calls.

fvanham-coveo:
Integrators can use `coveoua('set','custom',{context_website:'myweb'})` to pass a [context_website to every collect call](https://docs.coveo.com/en/l2bf0354/coveo-for-commerce/initialize-the-coveo-analytics-javascript). That context website will be injected on the root of the collect payload. For non-collect events that are sent using coveoua, the contents of this object would be passed as a `custom` object, whereas these events are expecting this content under `customData`. This is causing every integrator to have to manually ensure the `context_website` parameter is set correctly, on every event, by passing it explicitly. Missing one causes severe impact on metrics. See [discuss link here](https://discuss.coveo.com/t/logging-ua-events-customdata-vs-custom/8820).

This change modifies the behavior of coveoua to set these custom parameters on the `customData` object instead, for non-collect events. If an identical parameter already existed on the payload's `customData`, its value is not modified. If the value of the set `custom` object is `null` or `undefined`, a new object is not created.

The irony is that the docs and the programmed intent do not align: The proper and simplest way to set this parameters would've been `coveoua('set','context_website', '<value>')` not `coveoua('set','custom', {context_website:<value>})` 

**Notes:**
- Tests have been added to verify.
- In retrospect, I think it's a bad idea to allow injection of arbitrary custom parameters on the root payload of the collect event itself, because it will hamper validation and make it harder to process a collect event effectively. I did not change that behavior because it's already out and I've seen it in the wild.
- Plugin loading mocks were removed. It's possible to test initialization of a plugin by calling a specific plugin action and observing the results. Also, logic of `coveoua` depends on loaded plugins, which should reflect in the tests for `simpleAnalytics`.
- I changed the name `handleOneAnalyticsEvent` to `coveoua` for consistency with external usage and other tests.
- Fixed a minor issue with a plugin test.

### DIFF
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -30,6 +30,7 @@ import {isApiKey} from './token';
 import {isReactNative, ReactNativeRuntimeWarning} from '../react-native/react-native-utils';
 import {doNotTrack} from '../donottrack';
 import {NullStorage} from '../storage';
+import {isObject} from './utils';
 
 export const Version = 'v15';
 
@@ -433,7 +434,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
     private processCustomParameters(payload: IRequestPayload): IRequestPayload {
         const {custom, ...rest} = payload;
         let lowercasedCustom = {};
-        if (custom && typeof custom === 'object' && !Array.isArray(custom)) {
+        if (custom && isObject(custom)) {
             lowercasedCustom = this.lowercaseKeys(custom);
         }
 
@@ -447,7 +448,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
 
     private mapCustomParametersToCustomData(payload: IRequestPayload): IRequestPayload {
         const {custom, ...rest} = payload;
-        if (custom && typeof custom === 'object' && !Array.isArray(custom)) {
+        if (custom && isObject(custom)) {
             const lowercasedCustom = this.lowercaseKeys(custom);
             return {...rest, customData: {...lowercasedCustom, ...payload.customData}};
         } else {
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -269,7 +269,9 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
         const removeUnknownParameters: ProcessPayloadStep = (currentPayload) =>
             usesMeasurementProtocol ? this.removeUnknownParameters(currentPayload) : currentPayload;
         const processCustomParameters: ProcessPayloadStep = (currentPayload) =>
-            usesMeasurementProtocol ? this.processCustomParameters(currentPayload) : currentPayload;
+            usesMeasurementProtocol
+                ? this.processCustomParameters(currentPayload)
+                : this.mapCustomParametersToCustomData(currentPayload);
 
         const payloadToSend = await [
             cleanPayloadStep,
@@ -439,6 +441,16 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
             ...lowercasedCustom,
             ...newPayload,
         };
+    }
+
+    private mapCustomParametersToCustomData(payload: IRequestPayload): IRequestPayload {
+        const {custom, ...rest} = payload;
+        if (custom) {
+            const lowercasedCustom = this.lowercaseKeys(custom);
+            return {...rest, customData: {...lowercasedCustom, ...payload.customData}};
+        } else {
+            return payload;
+        }
     }
 
     private lowercaseKeys(custom: any) {
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -432,8 +432,10 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
 
     private processCustomParameters(payload: IRequestPayload): IRequestPayload {
         const {custom, ...rest} = payload;
-
-        const lowercasedCustom = this.lowercaseKeys(custom);
+        let lowercasedCustom = {};
+        if (custom && typeof custom === 'object' && !Array.isArray(custom)) {
+            lowercasedCustom = this.lowercaseKeys(custom);
+        }
 
         const newPayload = convertCustomMeasurementProtocolKeys(rest);
 
@@ -445,7 +447,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
 
     private mapCustomParametersToCustomData(payload: IRequestPayload): IRequestPayload {
         const {custom, ...rest} = payload;
-        if (custom) {
+        if (custom && typeof custom === 'object' && !Array.isArray(custom)) {
             const lowercasedCustom = this.lowercaseKeys(custom);
             return {...rest, customData: {...lowercasedCustom, ...payload.customData}};
         } else {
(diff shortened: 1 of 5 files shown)


src/coveoua/simpleanalytics.spec.ts
```diff
@@ -67,170 +50,162 @@ describe('simpleanalytics', () => {
...
         });
 
         it('uses EC and SVC plugins by default', () => {
-            handleOneAnalyticsEvent('init', 'SOME TOKEN');
-
-            expect(SVC).toHaveBeenCalled();
-            expect(EC).toHaveBeenCalled();
+            coveoua('init', 'SOME TOKEN');
+            expect(() => coveoua('callPlugin', 'ec', 'nosuchfunction')).toThrow(/does not exist/);
+            expect(() => coveoua('callPlugin', 'svc', 'nosuchfunction')).toThrow(/does not exist/);
```
samisayegh, line 129: Hmm, I'm not sure about these changes. It feels like we've changed the tests?
The mocks were removed, and now we're checking that errors are being thrown?

Whereas before we would assert whether one or both plugins had been invoked,

src/coveoua/simpleanalytics.spec.ts
```diff
@@ -239,142 +214,187 @@ describe('simpleanalytics', () => {
...
+            coveoua('set', 'custom', undefined);
+            await coveoua('send', 'view', {somedata: 'something'});
+
+            expect(fetchMock.calls().length).toBe(1);
+            let result = JSON.parse(fetchMock.lastCall()![1]!.body!.toString());
+            expect(result).toHaveProperty('somedata', 'something');
+            expect(result).not.toHaveProperty('customData');
+        });
+
+        it('will not override hardcoded custom parameters', async () => {
```
samisayegh, line 281: Would this test description be clearer if it said `... hardcoded customData parameters`?

src/coveoua/simpleanalytics.spec.ts
```diff
@@ -239,142 +214,187 @@ describe('simpleanalytics', () => {
...
+            coveoua('set', 'custom', null);
+            await coveoua('send', 'view', {somedata: 'something'});
+
+            expect(fetchMock.calls().length).toBe(1);
+            let result = JSON.parse(fetchMock.lastCall()![1]!.body!.toString());
+            expect(result).toHaveProperty('somedata', 'something');
+            expect(result).not.toHaveProperty('customData');
+        });
+
+        it('does not add a customData entry for custom params which are undefined', async () => {
```
samisayegh, line 270: Should we add a test for when `custom` is set to e.g. a string? i.e. checking that the value of `custom` is a truthy object?

src/coveoua/simpleanalytics.spec.ts
```diff
@@ -239,142 +214,187 @@ describe('simpleanalytics', () => {
...
+            await coveoua('send', 'pageview');
+
+            expect(fetchMock.calls().length).toBe(1);
+            let result = JSON.parse(fetchMock.lastCall()![1]!.body!.toString());
+            expect(result).toHaveProperty('context_website', 'MY_WEBSITE');
+        });
+
+        it('can set a custom_website parameter on a non-collect event', async () => {
+            coveoua('init', 'MYTOKEN', {plugins: []});
+            coveoua('set', 'custom', {context_website: 'MY_WEBSITE'});
```
samisayegh, line 250: Should we add a test checking that uppercase characters in the payload keys are being lowercased?

samisayegh (approved):
I'm less sure about the removal of the mocked plugins and changing the test assertions. That aside, making `collect` custom payloads compatible with legacy UA looks great.

src/client/analytics.ts
```diff
@@ -441,6 +443,16 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         };
     }
 
+    private mapCustomParametersToCustomData(payload: IRequestPayload): IRequestPayload {
+        const {custom, ...rest} = payload;
```
olamothe, line 447: What if instead of only sniping the key `custom`, we did the opposite ?

Every key that *should* be top level gets filtered out, and `...rest` gets mapped to `customData`.

To answer this concern here: 
```
In retrospect, I think it's a bad idea to allow injection of arbitrary custom parameters on the root payload of the collect event itself, because it will hamper validation and make it harder to process a collect event effectively. I did not change that behavior because it's already out and I've seen it in the wild.
```

src/client/analytics.ts
```diff
@@ -441,6 +443,16 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        if (custom) {
+            const lowercasedCustom = this.lowercaseKeys(custom);
```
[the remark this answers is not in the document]
fvanham-coveo: The behavior in that case is undefined, so I'm not even sure if the above is an unexpected overwrite. If we lowercase keys, there's always the potential for multiple values mapping to one key. I did not change existing behavior here.

src/coveoua/simpleanalytics.spec.ts
fvanham-coveo: The mocks were removed, because they were only needed for one set of particular tests (the ones above). It checked the plugin was loaded by stubbing with an empty plugin and intercepting the loader. The behavior of subsequent tests was then sometimes incorrect, because it depended on specific plugins being loaded. 

The initial tests test whether the loader for the plugin had been invoked. The current tests test further downstream by attempting to call a plugin with a non-existent function. It's useful to test whether the plugin was loaded. I'm generally not in favor of testing-by-exception but didn't see another way without messing up the js loaders.

src/coveoua/simpleanalytics.spec.ts
fvanham-coveo: Well, half the code said they are set on `custom` the other half said they should be on `customData` :) I'm ok with anything.

src/client/analytics.ts
fvanham-coveo: There is no `customData` on collect. So I'm not sure what that option would imply. We add a layered payload on top of a flat protocol? Further butcher collect :)? I'm not against it parse, but it's a way bigger change than I was trying to make here.

src/coveoua/simpleanalytics.spec.ts
fvanham-coveo: These exist in `/functional`

src/coveoua/simpleanalytics.spec.ts
fvanham-coveo: Yeah, you're right. Will add.

fvanham-coveo: I added tests for this case. The behavior was quite bad in edge cases (e.g. passing string `test` on `custom` would result in entries `0:t`, `1:e`, `2:s` and `3:t` being added to the collect payload.

olamothe (approved):

src/coveoua/simpleanalytics.spec.ts
samisayegh: Nice!

src/client/analytics.ts
```diff
@@ -430,8 +432,10 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
 
     private processCustomParameters(payload: IRequestPayload): IRequestPayload {
         const {custom, ...rest} = payload;
-
-        const lowercasedCustom = this.lowercaseKeys(custom);
+        let lowercasedCustom = {};
+        if (custom && typeof custom === 'object' && !Array.isArray(custom)) {
```
samisayegh, line 436: What do you think of extracting this logic into a function e.g. `isObject`?

fvanham-coveo: Yeah I can do that.

PR #391, opened 2023-01-06
feat(insight): added user actions events

nathanlb:
Added the following event logging methods related to the Quantic Insight Panel:
- openUserActions
- showPrecedingSessions
- showFollowingSessions
- logViewedDocumentClick
- logPageViewClick

### DIFF
--- a/src/insight/insightClient.ts
+++ b/src/insight/insightClient.ts
@@ -24,7 +24,6 @@ import {
     InsightPagerMetadata,
     InsightResultsSortMetadata,
     UserActionsDocumentMetadata,
-    UserActionsPageViewMetadata,
 } from './insightEvents';
 
 export interface InsightClientProvider {
@@ -240,13 +239,6 @@ export class CoveoInsightClient {
         return this.logCustomEvent(InsightEvents.clickViewedDocument, {
             ...generateMetadataToSend(metadata, false),
             document,
-        });
-    }
-
-    public logPageViewClick(pageView: UserActionsPageViewMetadata, metadata: CaseMetadata) {
-        return this.logCustomEvent(InsightEvents.clickPageView, {
-            ...generateMetadataToSend(metadata, false),
-            pageView,
         });
     }
 
--- a/src/insight/insightClient.ts
+++ b/src/insight/insightClient.ts
@@ -23,6 +23,8 @@ import {
     InsightQueryErrorMeta,
     InsightPagerMetadata,
     InsightResultsSortMetadata,
+    UserActionsDocumentMetadata,
+    UserActionsPageViewMetadata,
 } from './insightEvents';
 
 export interface InsightClientProvider {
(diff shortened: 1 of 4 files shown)


mmitiche (approved):
LGTM 👌

nathanlb:
> Please bump patch version in `package.json`.

Oh my bad I assumed it was done automatically like most of our other open source projects. Is there a reason it isn't? 🤔

src/searchPage/searchPageEvents.ts
```diff
@@ -322,6 +322,9 @@ export const CustomEventsTypes: Partial<Record<SearchPageEvents | InsightEvents,
     [SearchPageEvents.clearRecentResults]: 'recentlyClickedDocuments',
     [SearchPageEvents.showLessFoldedResults]: 'folding',
     [InsightEvents.expandToFullUI]: 'interface',
+    [InsightEvents.openUserActions]: 'User Actions',
+    [InsightEvents.showPrecedingSessions]: 'User Actions',
+    [InsightEvents.showFollowingSessions]: 'User Actions',
```
lbergeron, line 327: The casing is different from other event types. Could the value be `userActions` to follow the same format?

nathanlb: Yeah I agree it's weird. I did it like this because "User Actions" is the event type currently sent by the JSUI insight panel for `openUserActions`. For the analytics events to be compatible between the JSUI and Quantic version I think we would need to keep them the same.

lbergeron: Makes sense!

fvanham-coveo:
> > Please bump patch version in `package.json`.
> 
> Oh my bad I assumed it was done automatically like most of our other open source projects. Is there a reason it isn't? 🤔

Disregard my comment. There is an automated version bump by Jenkins. 🤦

fvanham-coveo (approved):

src/insight/insightClient.ts
```diff
@@ -219,6 +221,35 @@ export class CoveoInsightClient {
...
+    }
+
+    public logViewedDocumentClick(document: UserActionsDocumentMetadata, metadata: CaseMetadata) {
+        return this.logCustomEvent(InsightEvents.clickViewedDocument, {
+            ...generateMetadataToSend(metadata, false),
+            document,
+        });
+    }
+
+    public logPageViewClick(pageView: UserActionsPageViewMetadata, metadata: CaseMetadata) {
```
mmitiche, line 246: Is this supposed to be logged when the user clicks on a user action of type view in the timeline?

src/insight/insightEvents.ts
```diff
@@ -20,6 +20,26 @@ export enum InsightEvents {
...
+    /**
+     * Identifies the event that gets logged when the user clicks to view user action sessions preceding the ticket creation.
+     */
+    showPrecedingSessions = 'showPrecedingSessions',
+    /**
+     * Identifies the event that gets logged when the user clicks to view user action sessions following the ticket creation.
+     */
+    showFollowingSessions = 'showFollowingSessions',
+    /**
+     * Identified the event that gets logged when the user clicks a viewed document.
```
lbergeron, line 36:
```suggestion
     * Identifies the event that gets logged when the user clicks a viewed document.
```

src/insight/insightEvents.ts
```diff
@@ -20,6 +20,26 @@ export enum InsightEvents {
...
+     */
+    clickViewedDocument = 'clickViewedDocument',
+    /**
+     * Identified the event that gets logged when the user clicks a viewed page.
```
lbergeron, line 40:
```suggestion
     * Identifies the event that gets logged when the user clicks a viewed page.
```

src/insight/insightClient.ts
nathanlb: Yes that's the idea

lbergeron (approved):

src/insight/insightClient.ts
mmitiche: The reason I'm asking is because in the standard agent panel, the user actions of type `view` in the timeline are not clickable:
StandardUA
I asked UX about this, [UX discussion ](https://www.figma.com/file/tc1NDbxceu3tW1UjFIt8cm?node-id=207:15790#341585226), and it seems that we wanna keep the same behaviour.

IMO, I find it very useful to have the user action of type `view` clickable so I just want to confirm our position regarding this.

nathanlb: Ah ok it's unfortunate that the mockup was not updated because it still shows them as links and that's what I went off of. 😅
Screen Shot 2023-01-11 at 1 47 12 PM

I'll remove this event then, no problem

mmitiche: @erocheleau  do you have more insight regarding this? do we have some particular reason why we didn't make the user action of type `view` clickable in the standard agent panel?

nathanlb: Looks to me like it can be clickable if we have the right information no? 
https://github.com/coveo/search-ui-extensions/blob/0039882563f36f0fa21f751c769d36ea902816df/src/components/UserActions/UserActivity.ts#L454

mmitiche: yes true! so I think we do need the analytics action, sorry about that 😅

nathanlb: Restored it. 😛 No worries I was a bit too quick to pull the trigger too.

erocheleau: Yea afaik, not all views will be a uri we have, some will just be a permanentid in the contentidkey/value pair so that will be the only data we have, but if we get a url I think in the `contentidvalue` data, we should display as a link?

I wouldn't say it's a "requirement" however.

mmitiche: @erocheleau What do you mean by "not all views will be a uri we have"? what are the uri that we have?

erocheleau: No all pageviews events will contain a uri in the data

nathanlb: Sooooo do you guys think the metadata for this event is suitable?

erocheleau: Don't think we can really do better, this will contain the case data + the title/contentidkey/contentidvalue right?

nathanlb: exactly

PR #392, opened 2023-01-13
[UA-7290] Fix tracking issues

fvanham-coveo:
After closer investigation of the cookie generation code I found a few bugs:
- The first party cookie setting code did not do what it says it does. For a domain `some.domain.blah.com` it would try to write the cookie to `.domain.blah.com`. This is not wrong, but better would be to write to `blah.com`.
- The first party cookie was being set without expiration date, meaning it effectively acts **as a session cookie**. This means that cross sub-domain tracking won't work, as the cookie would expire at the end of a session.

In rare cases where a consistent id is needed across domains, I've provided an api call which allows a user to set a custom uuid, either directly from an existing uuid or from a given string. Coveo doesn't provide cross domain tracking in its code. If this string option is used, a cross-domain-consistent namespace string also needs to be provided, to increase entropy. Note that plaintext strings are not acceptable as clientIds, as they may (and have in the past) contain unintended personal information.

I'm relying on the industry standard `uuid` library to validate and generate uuid.v5. There is still code in the library which generates uuids v4. We should remove this in another change, as it's relying on `random()` to polyfill the browsers crypto.getRandomValues(). This in itself carries a risk to generate potentially overlapping uuids. At this point any browser [which does not support](https://caniuse.com/getrandomvalues) `getRandomValues` is either a decade old, or is called "Opera Mini". This additional dependency has increased the library size from 88kb to 91kb.

Small edits:
- Removed console log in tests by checking for presence of `global` variable.
- Fixed missing cross-fetch dependency warning in builds.

I would suggest everyone reviewing this to checkout the branch, run `npm run start`, set up two localhosts aliases (via /etc/hosts) and verify.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -7,11 +7,6 @@
         O=o.getElementsByTagName(v)[0];O.parentNode.insertBefore(u,O)
         })(window,document,'script','coveoua.debug.js')
 
-        // setting a consistent clientId
-        coveoua('onLoad', function (args) {
-            coveoanalytics.getCurrentClient().setClientId('myclientid', 'mynamespace');
-        });
-
         // Replace YOUR-TOKEN with your real token
         // (eg: an API key which has the rights to write into Coveo UsageAnalytics)
         coveoua('init', 'YOUR-TOKEN');
--- a/public/index.html
+++ b/public/index.html
@@ -7,6 +7,11 @@
         O=o.getElementsByTagName(v)[0];O.parentNode.insertBefore(u,O)
         })(window,document,'script','coveoua.debug.js')
 
+        // setting a consistent clientId
+        coveoua('onLoad', function (args) {
+            coveoanalytics.getCurrentClient().setClientId('myclientid', 'mynamespace');
+        });
+
         // Replace YOUR-TOKEN with your real token
         // (eg: an API key which has the rights to write into Coveo UsageAnalytics)
         coveoua('init', 'YOUR-TOKEN');
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -22,12 +22,6 @@ const tsPlugin = () =>
         useTsconfigDeclarationDir: true,
     });
 
-const nodeResolvePlugin = () =>
-    nodeResolve({
-        preferBuiltins: true,
-        only: ['cross-fetch', 'uuid'],
-    });
-
 const browserUMD = {
     input: './src/coveoua/browser.ts',
     output: [
@@ -54,7 +48,7 @@ const browserUMD = {
     ],
     plugins: [
         browserFetch(),
-        nodeResolvePlugin(),
+        nodeResolve({preferBuiltins: true, only: ['uuid']}),
         tsPlugin(),
         process.env.SERVE
             ? serve({
@@ -75,7 +69,12 @@ const nodeCJS = {
         file: './dist/library.js',
         format: 'cjs',
     },
-    plugins: [nodeResolve({mainFields: ['main'], preferBuiltins: true}), commonjs(), tsPlugin(), json()],
+    plugins: [
+        nodeResolve({mainFields: ['main'], preferBuiltins: true, only: ['uuid']}),
+        commonjs(),
+        tsPlugin(),
+        json(),
+    ],
 };
 
 const browserESM = {
@@ -86,7 +85,7 @@ const browserESM = {
     },
     plugins: [
         browserFetch(),
-        nodeResolvePlugin(),
+        nodeResolve({preferBuiltins: true, only: ['uuid']}),
         typescript({
             useTsconfigDeclarationDir: true,
             tsconfigOverride: {compilerOptions: {target: 'es6'}},
@@ -101,7 +100,7 @@ const libRN = {
         format: 'es',
     },
     plugins: [
-        nodeResolvePlugin(),
+        nodeResolve({preferBuiltins: true, only: ['uuid']}),
         commonjs(),
         json(),
         typescript({
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -22,6 +22,12 @@ const tsPlugin = () =>
         useTsconfigDeclarationDir: true,
     });
 
+const nodeResolvePlugin = () =>
+    nodeResolve({
+        preferBuiltins: true,
+        only: ['cross-fetch', 'uuid'],
+    });
+
 const browserUMD = {
     input: './src/coveoua/browser.ts',
     output: [
@@ -48,6 +54,7 @@ const browserUMD = {
     ],
     plugins: [
         browserFetch(),
+        nodeResolvePlugin(),
         tsPlugin(),
         process.env.SERVE
             ? serve({
@@ -68,7 +75,7 @@ const nodeCJS = {
         file: './dist/library.js',
         format: 'cjs',
     },
-    plugins: [nodeResolve({mainFields: ['main']}), commonjs(), tsPlugin(), json()],
+    plugins: [nodeResolve({mainFields: ['main'], preferBuiltins: true}), commonjs(), tsPlugin(), json()],
 };
 
 const browserESM = {
@@ -79,6 +86,7 @@ const browserESM = {
     },
     plugins: [
         browserFetch(),
+        nodeResolvePlugin(),
         typescript({
             useTsconfigDeclarationDir: true,
             tsconfigOverride: {compilerOptions: {target: 'es6'}},
@@ -93,6 +101,9 @@ const libRN = {
         format: 'es',
     },
     plugins: [
+        nodeResolvePlugin(),
+        commonjs(),
+        json(),
         typescript({
             useTsconfigDeclarationDir: true,
             tsconfigOverride: {compilerOptions: {target: 'es6'}},
--- a/src/client/analytics.spec.ts
+++ b/src/client/analytics.spec.ts
@@ -453,3 +453,34 @@ describe('doNotTrack', () => {
         expect(Cookie.get('coveo_visitorId')).not.toBe(aVisitorId);
     });
 });
+
+describe('custom clientId', () => {
+    it('allows setting of a custom clientId', async () => {
+        let client = new CoveoAnalyticsClient({});
+        client.setClientId('c7d57b22-4aa8-487a-a106-be5243885f0a');
+        expect(await client.getCurrentVisitorId()).toBe('c7d57b22-4aa8-487a-a106-be5243885f0a');
+    });
+
+    it('allows setting a custom consistent clientId given a string', async () => {
+        let client = new CoveoAnalyticsClient({});
+        client.setClientId('somestring', 'testNameSpace');
+        //uuid v5 specific uuid generation
+        expect(await client.getCurrentVisitorId()).toBe('2c356915-8223-5773-acb8-e2a34404a559');
+        //check for consistent ids
+        client.setClientId('somestring', 'testNameSpace');
+        expect(await client.getCurrentVisitorId()).toBe('2c356915-8223-5773-acb8-e2a34404a559');
+        client.setClientId('otherstring', 'testNameSpace');
+        expect(await client.getCurrentVisitorId()).not.toBe('2c356915-8223-5773-acb8-e2a34404a559');
+        client.setClientId('somestring', 'otherNameSpace');
+        expect(await client.getCurrentVisitorId()).not.toBe('2c356915-8223-5773-acb8-e2a34404a559');
+    });
+
+    it('errors when not providing a namespace', async () => {
+        let client = new CoveoAnalyticsClient({});
+        expect.assertions(1);
+        await expect(client.setClientId('somestring')).rejects.toEqual(
+            Error('Cannot generate uuid client id without a specific namespace string.')
+        );
+        //uuid v5 specific uuid generation
+    });
+});
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -23,6 +23,7 @@ import {hasWindow, hasDocument} from '../detector';
 import {addDefaultValues} from '../hook/addDefaultValues';
 import {enhanceViewEvent} from '../hook/enhanceViewEvent';
 import {uuidv4} from './crypto';
+import {v5 as uuidv5, validate as uuidValidate} from 'uuid';
 import {
     convertKeysToMeasurementProtocol,
     isMeasurementProtocolKey,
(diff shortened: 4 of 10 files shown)


jmprovencher:
QUestion out of scope for this PR. In the latest version of this library, is there still anything sending a visitorId anywhere?

public/index.html
```diff
@@ -7,6 +7,11 @@
         O=o.getElementsByTagName(v)[0];O.parentNode.insertBefore(u,O)
         })(window,document,'script','coveoua.debug.js')
 
+        // setting a consistent clientId
+        coveoua('onLoad', function (args) {
+            coveoanalytics.getCurrentClient().setClientId('myclientid', 'mynamespace');
+        });
```
jmprovencher, line 13: what is this for ?

src/client/analytics.spec.ts
```diff
@@ -453,3 +453,34 @@ describe('doNotTrack', () => {
...
+        let client = new CoveoAnalyticsClient({});
+        client.setClientId('c7d57b22-4aa8-487a-a106-be5243885f0a');
+        expect(await client.getCurrentVisitorId()).toBe('c7d57b22-4aa8-487a-a106-be5243885f0a');
+    });
+
+    it('allows setting a custom consistent clientId given a string', async () => {
+        let client = new CoveoAnalyticsClient({});
+        client.setClientId('somestring', 'testNameSpace');
+        //uuid v5 specific uuid generation
+        expect(await client.getCurrentVisitorId()).toBe('2c356915-8223-5773-acb8-e2a34404a559');
```
jmprovencher, line 468: would it be possible (maybe in another PR) to completely refactor any reference to visitorID -> to be changed to -> clientId ?

jmprovencher: or is this breaking ?

src/client/analytics.ts
```diff
@@ -118,6 +119,8 @@ export function buildBaseUrl(endpoint = Endpoints.default, apiVersion = Version)
     return `${endpoint}${endpointIsCoveoProxy ? '' : '/rest'}/${apiVersion}`;
 }
 
+const COVEO_NAMESPACE = '38824e1f-37f5-42d3-8372-a4b8fa9df946'; //note: changing this value will destroy the mapping from tracking string to clientId.
+
```
jmprovencher, line 123: This means if anybody changes that string, it breaks all existing tracking for all customers? If so, should the explanation be longer and possibly more clear that this is a big deal?

jmprovencher (approved):

jmprovencher:
So if I understand correctly, this PR would fix the following problem.

**Before changes**

_products.dell.com. -> cookie a 

service.dell.com  -> cookie b_


**After changes**

_products.dell.com. -> cookie a 

service.dell.com  -> cookie a_

src/client/analytics.ts
```diff
@@ -209,6 +212,15 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         await this.storage.setItem('visitorId', visitorId);
     }
 
+    async setClientId(value: string, namespace?: string) {
```
lbourdages, line 215: If this is not used, it will still generate a random (uuidv4) uuid?

lbourdages (approved):
I'm not a TS expert but looks good to me. One thing to note is that, if I understand correctly, bumping to this version would reset all IDs because the cookie domain changes (e.g. from shop.dell.com to dell.com). While I think it is acceptable if it fixes things (and enables cross-site tracking), it's to keep in mind.

src/cookieutils.ts
```diff
@@ -67,7 +43,10 @@ export class Cookie {
     }
 }
 
-function writeCookie(details: CookieDetails) {
-    const {name, value, expires, domain} = details;
-    document.cookie = `${name}=${value}${expires}; path=/; domain=${domain}; SameSite=Lax`;
+function writeCookie(name: string, value: string, expirationDate?: Date, domain?: string) {
+    document.cookie =
+        `${name}=${value}` +
+        (expirationDate ? `;expires=${(expirationDate as any).toGMTString()}` : '') +
```
olamothe, line 49: You can use `toUTCString`, which will remove the need for the `any` cast.

Explanation: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toUTCString#aliasing

```suggestion
        (expirationDate ? `;expires=${expirationDate.toUTCString()}` : '') +
```

src/client/analytics.spec.ts
olamothe: It would be. 

While I think the amount of clients that would actually reference those function directly is low, these functions are still globally exported and could start breaking random integration on upgrade.

I would propose a transition period on v2 where: 

1- Create new functions/variables that uses the proper naming for `clientId`
2- Alias all the old ones so that they just call the new function names/use new property name under the hood, and log a warning anytime the old function/property is used. The warning provides clear instruction on how to upgrade.
3- Plan for eventual v3 and remove everything that's marked deprecated.

src/client/analytics.ts
olamothe, line 215: Previously there was no code exported that would allow an implementer to provide their own clientID, as those ID were all internally generated.

And code would check if the ID exists before generating a new one on initialization.

 It means it was ~impossible~ harder for an implementer to fudge their implementation and, for example, incorrectly "wipe" their ID and regenerate new ones on every new end user visit.

With this new method, do you think we should add any kind of mechanism to prevent (or lessen the chances) of an implementer doing this ?

olamothe (approved):

src/client/analytics.spec.ts
fvanham-coveo: Yes, I had the same thought.

public/index.html
fvanham-coveo: This is a dev oriented test script, allowing you to 

> So if I understand correctly, this PR would fix the following problem.
> 
> **Before changes**
> 
> _products.dell.com. -> cookie a_
> 
> _service.dell.com -> cookie b_
> 
> **After changes**
> 
> _products.dell.com. -> cookie a_
> 
> _service.dell.com -> cookie a_

No, before change should be

_products.dell.com -> session cookie a, which expired after the session_

_service.dell.com -> session cookie a, which expired after the session_

src/client/analytics.ts
fvanham-coveo: Yes. That code is is in `determineVisitorId`. The first party cookie domain does change, but this was a session cookie so there's no long term effect.

fvanham-coveo: This new method should ONLY be used by implementers that want to set their own client ids. It's not mandatory, and only intended to serve those cases where client already has a pseudo client id available. If it's not called, we generate and store one ourselves as used to be the case. I can flag cases where the clientId changes with a console error, but depending on the chosen client implementation this may be a valid use: switching a logged in user and basing the client id off of user id for example. Maybe this is a case for removing the line from index.html, since it may give off the wrong impression.

fvanham-coveo:
> QUestion out of scope for this PR. In the latest version of this library, is there still anything sending a visitorId anywhere?

Yes, visitorId is tacked onto the url in all calls to UA endpoints.

fvanham-coveo:
> So if I understand correctly, this PR would fix the following problem.
> 
> **Before changes**
> 
> _products.dell.com. -> cookie a_
> 
> _service.dell.com -> cookie b_
> 
> **After changes**
> 
> _products.dell.com. -> cookie a_
> 
> _service.dell.com -> cookie a_

The initial situation was slightly different: both sites would have cookie a, but with session-level expiration.

src/cookieutils.ts
```diff
@@ -1,50 +1,27 @@
...
-
-            // check if cookie was successfuly set to the given domain
-            // (otherwise it was a Top-Level Domain)
-            if (Cookie.get(name) == null || Cookie.get(name) != value) {
-                // append "." to current domain
-                domain = '.' + host;
-                writeCookie({name, value, expires, domain});
-            }
+            // we always have at least 2 domain parts
+            domain = domainParts[domainParts.length - 2] + '.' + domainParts[domainParts.length - 1];
```
rphilippen, line 23: This is a personal preference, but I would refactor this code to define a module local `topLevelDomain` function that extracts the top level domain from a hostname. That would change this code into:

```typescript
        let hostDomain = topLevelDomain(window.location.hostname);
        if (hostDomain.contains('.')) {
            writeCookie(name, value, expirationDate, hostDomain);
        } else {
            // no "." in a domain - single domain name, it's localhost or something similar
            writeCookie(name, value, expirationDate);
        }
```

For the function itself, the idiomatic way of writing that in TypeScript/JavaScript would be this:
```typescript
/**
 * Return the top level domain from an input.
 * Examples:
 *     "coveo.com" -> "coveo.com"
 *     "some.example.coveo.com" -> "coveo.com"
 *     "localhost" -> "localhost"
 * @param hostname The hostname to get the top level domain name from.
 * @returns The top level domain.
 */
function topLevelDomain(hostname) {
    return hostname.split('.').slice(-2).join('.');
}
```

If you want to be less wasteful on allocating intermediate Arrays (for `split` and `slice`), you could write it like this:
```typescript
/**
 * Return the top level domain from an input.
 * Examples:
 *     "coveo.com" -> "coveo.com"
 *     "some.example.coveo.com" -> "coveo.com"
 *     "localhost" -> "localhost"
 * @param hostname The hostname to get the top level domain name from.
 * @returns The top level domain.
 */
function topLevelDomain(hostname: string): string {
    let i = hostname.lastIndexOf('.');
    i = i > 0 ? hostname.lastIndexOf('.', i - 1) : -1;
    return hostname.substring(i + 1)
}
```

rphilippen (approved):

rollup.config.js
```diff
@@ -48,6 +54,7 @@ const browserUMD = {
     ],
     plugins: [
         browserFetch(),
+        nodeResolvePlugin(),
```
samisayegh, line 57: Are we sure we should be using this plugin in the browser bundle? e.g.
I see the node plugin mentions `cross-fetch` in the array on line 28, but there is already a `browserFetch` plugin here on line 56. It seems like we're bundling extra code?

Similiarly for `uuid`, if that dependency relies on run-time globals, I'd image we'd want to pull in the browser file in the case of the browser bundle.

src/cookieutils.ts
```diff
@@ -53,8 +30,7 @@ export class Cookie {
         var cookieArray = document.cookie.split(';');
         for (var i = 0; i < cookieArray.length; i++) {
             var cookie = cookieArray[i];
-            cookie = cookie.replace(/^\s+/, '');
-
+            cookie = cookie.replace(/^\s+/, ''); //strip whitespace from front of cookie only
             if (cookie.indexOf(cookiePrefix) == 0) {
```
rphilippen, line 34: This code has written out [startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith), but poorly. The "poor man's version" should use `.lastIndexOf(prefix, 0)` to avoid searching the whole string for no good reason at all. However, since support for `startsWith` is so ubiquitous, I suggest:
```suggestion
            if (cookie.startsWith(cookiePrefix)) {
```

rollup.config.js
fvanham-coveo: You're right, for browser fetch. That should not end up here, for the browser distributable I think we're relying on the native fetch implementation. uuid is a npm package which is external, so needs to be included. I'll customize the imports for each distributable.

jmprovencher:
> > QUestion out of scope for this PR. In the latest version of this library, is there still anything sending a visitorId anywhere?
> 
> Yes, visitorId is tacked onto the url in all calls to UA endpoints.

Do we have a story to stop using visitorId in UA endpoints ? I thought this was already done months ago. Client id was suppose to only be used with newer version of the library??

src/client/analytics.ts
samisayegh: Hmm, it feels risky to make changing the `clientId` easier. Using your example, I could imagine a faulty implementation where a customer supplies the same user id for many users. All of a sudden we see many events belonging to the same client id, when in fact they belong to multiple users on different computers.

Curious to know the use-case at play?

If it's for stitching, we could separate the concerns? The customer's id can be sent as metadata. This gives them the freedom to use their format which may differ from uuid too. The `clientId` remains an internal Coveo-specific identifier,

If the idea is to identify the _same_ user across multiple devices, I still wonder if we should separate the two? Separating would a) ensure a base-level of functionality, and b) having a dedicated method `setUserId` allows us to explicitly document how the different ids behave. e.g.

- `clientId`: identifies a browser client. Different users using the same browser will share the same clientId. The same user using multiple browsers on a single device, or using the same browser on multiple devices will have different client ids. Coveo manages this id via it's `coveo.analytics` library.
- `userId`: identifies specific user across multiple browsers and devices. Coveo does not manage this id. Customers are responsible for ensuring this id is set correctly.


A disadvantage of separating the two is it introduces an if/else in downstream consuming code, but this might be useful in the case of e.g. ML, where they can tailor the strategy based, make different assumptions about the data based on the id available to them.

samisayegh (approved):
Main concern is around giving control over client ids. Otherwise looks good,

fvanham-coveo:
> > > QUestion out of scope for this PR. In the latest version of this library, is there still anything sending a visitorId anywhere?
> > 
> > 
> > Yes, visitorId is tacked onto the url in all calls to UA endpoints.
> 
> Do we have a story to stop using visitorId in UA endpoints ? I thought this was already done months ago. Client id was suppose to only be used with newer version of the library??

Let me verify my statement. I saw tests for the feature, so not sure it's removed.

src/client/analytics.ts
```diff
@@ -209,6 +212,15 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
...
+        if (uuidValidate(value)) {
+            this.setCurrentVisitorId(value);
+        } else {
+            if (!namespace) throw Error('Cannot generate uuid client id without a specific namespace string.');
```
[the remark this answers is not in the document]
rphilippen: I agree with your statement, I do think you accidentally replied to the wrong comment.

src/client/analytics.ts
fvanham-coveo: Different users using the same browser will share the same clientId: this is not exactly true, which is the reason for this extension. Different users using the same browser will share the same clientId **only on the same top level domain**. If a customer wants to track cross-site, they can supply their own tracking id. For anonymous users this will be exceedingly hard (and carries privacy concerns) but if they insist it's possible. For logged in users, this can be an implementation choice: either generate a self-created clientId based off of the userId or use the userId to do the stitching. The choice made has consequences for the reporting metrics and is up to the customer.

rphilippen (approved):

PR #393, opened 2023-01-20
[UA-7345] Adding version calls

fvanham-coveo:
As a diagnostics tool it would be helpful to be able to query the current version of coveo.analytics.js at runtime from the browser or integration package. This PR adds a public facing `coveoua('version')` command for browser usage, and exposes the version on `AnalyticsClient` as a variable for any other integrations. I will also ensure public doc gets updated.

Note: I fixed some syntactic sugar to placate TypeScript on the side.


package.json
```diff
@@ -54,6 +54,7 @@
         "rollup-plugin-serve": "^1.0.1",
         "rollup-plugin-terser": "^7.0.2",
         "rollup-plugin-typescript2": "^0.27.0",
+        "rollup-plugin-version-injector": "^1.3.3",
```
samisayegh, line 57: An alternative could be to use the [replace](https://www.npmjs.com/package/@rollup/plugin-replace) plugin. It's less specialized, making it possible to inject other values in the future if needed:

```
//rollup.config.js
import packageJson from '../package.json'

replace({'process.env.PKG_VERSION': packageJson.version})

// version.ts

export lib_version = process.env.PKG_VERSION
export lib_date = Date.now() // or iso string if readability is preferred
```

rollup.config.js
```diff
@@ -49,6 +50,7 @@ const browserUMD = {
     plugins: [
         browserFetch(),
         nodeResolve({preferBuiltins: true, only: ['uuid']}),
+        versionInjector({logLevel: 'warn'}),
```
samisayegh, line 53: If the configured plugin is stored in a variable, you could reference the variable in each bundle instead of duplicating the `warn` config. If in the future you want to adjust the config, you'd only need to update one place.

src/version.ts
```diff
@@ -0,0 +1,2 @@
+export const lib_version = '[VI]{version}[/VI]';
+export const lib_build_date = '[VI]{date}[/VI]';
```
samisayegh, line 2: It doesn't seem like `lib_build_date` is referenced in the PR. Maybe it can be removed? Or should we expose this information as well?

For the variable names, is there a reason for using snakecase vs. the camelcase convention in js projects?

samisayegh (approved):

src/version.ts
fvanham-coveo: I added this for future use, but I guess the version only should be enough here. I'll remove it.

src/coveoua/simpleanalytics.ts
```diff
@@ -137,6 +138,10 @@ export class CoveoUA {
         this.plugins = new Plugins();
         this.params = {};
     }
+
+    version(): string {
+        return `${libVersion}`;
```
rphilippen, line 143: Is there a reason to put this inside a template literal? I'd simply expect
```suggestion
        return libVersion;
```

rphilippen (approved):

src/coveoua/simpleanalytics.ts
fvanham-coveo: Leftover from a previous change :) I'll fix it.

PR #394, opened 2023-02-03
Remove crypto implementation

fvanham-coveo:
coveo.analytics.js has had its own crypto implementation backfill in crypto.ts based on `Math.random()` which was (likely) polyfilling for browsers that, at the time, did not have `hasCryptoRandomValues` available. I'm creating a PR to switch the uuid implementation to a more commonly used external package ([uuid](https://www.npmjs.com/package/uuid) has 83M weekly downloads). Rationale for this is as follows:
- UUID generation using Math.random [is highly discouraged](https://crypto.stackexchange.com/questions/81619/is-it-possible-to-create-a-128-bit-uuid-from-a-weak-entropy-source) due to the risk of collisions. UUID generation across different versions of the UUID standard is best left to standardized packages. 
- Most (if not all) modern browsers [support](https://caniuse.com/getrandomvalues) `hasCryptoRandomValues` as of 2016 at the latest, except Opera Mini (whose javascript support is limited anyway).
- We were already integrating the uuid package to support persistent uuids from strings. Makes sense to align on a single package for consistency.

Downside: clients running on opera mini will not get a uuid generated as the library will error out. I'm not even sure opera mini [supports local storage](https://caniuse.com/?search=localStorage), and [it doesn't support cookies](https://caniuse.com/?search=cookie), so we're kind of in trouble anyway there :)


src/plugins/ec.spec.ts
```diff
@@ -25,7 +26,7 @@ describe('EC plugin', () => {
     beforeEach(() => {
         jest.clearAllMocks();
         client = createAnalyticsClientMock();
-        ec = new ECPlugin({client, uuidGenerator: someUUIDGenerator});
+        ec = new ECPlugin({client});
```
samisayegh, line 29: It feels more transparent to inject a mocked dependency instead of relying on `jest.mock`, because this requires inspecting the module's implementation (and possibly it's base class if the import is only used at that level).

Not against using `jest.mock`, but making this change without adjusting the `ECPlugin` constructor feels inconsistent. Said differently, is there any code left specifying `uuidGenerator`? If not ... the constructor arg can be removed, and we might apply the same logic on the base class's arguments.

Just exploring the other side, but personally, I'd lean towards the dependency injection pattern for its transparency,

samisayegh (approved):

src/plugins/ec.spec.ts
fvanham-coveo: Yes, you're right. I removed them because typescript was complaining since the declared type for `uuidgenerator` is `typeof v4`. A function declaration would've been better. I reverted the change.

jmprovencher (approved):

olamothe (approved):
That's a great change 💯

olamothe:
As for opera mini, yeah, we don't have to worry too much about that 👍

PR #399, opened 2023-03-01
UA-7368 remove rollup-plugin-polyfill-node dependency

samisayegh:
**Context**

The rollup warnings were due to importing `URL` from the `url` module instead of referencing the global variable.
The `url` module is a nodejs module rather than a package.json dependency. When rollup created the browser bundle, it correctly detected the mismatch, and suggested adding node polyfills to the browser bundle.

**Solution**

This PR removes the import in favor of use the global (note that the other instance of `URL` in `analytics.ts` was already doing this). The global is available on [nodejs](https://nodejs.org/api/url.html#new-urlinput-base) and modern [browsers](https://caniuse.com/url).

**Aside 1**

Being transparent, I did not see any difference in bundle size with or without the plugin. In both cases, the `coveoua.browser` bundle was 94kb. Still I think there's a benefit in not introducing the dependency.

More interesting point though is the increase in the last few months. The last build on my machine was from Aug 5th, 2022. Comparison of the bundle sizes between then and now:

|file | Aug 2022 (kb)| Mar 2023 (kb)|
|---|---|---|
|coveoua.browser|61| 94
|coveoua|61|94
|library.es|81|122
|library|718|784
|react-native|6|708

Only worth focusing on the first three since these are the ones that run in the browser. The increase has been ~50% in the last 6 months. Maybe there are good reasons, but it feels suspicious? If the size continues to increase to the point where clients start complaining about it, we'd be in reactive rather than proactive mode, which is never fun.

Just want to put this on your radar for now, but some actions could include:
- Investigating what caused the recent increase
- Adding a github action that reports the increase in bundle size on PRs.
- Looking for optimizations and quick wins.

**Aside 2**
[Thought for a separate PR, curious to get your thoughts on this topic]

If we want `coveo.analytics.js` to be more resilient, and run in any JS runtime without crashing (e.g. [react-native]), it could be better to use a function to retrieve globals, that checks if they exist, and fails gracefully when they don't. e.g.

```
function getUrlGlobal() {
  return global.URL ? global.URL : null;
}

const URL = getUrlGlobal()

if (!URL) {
   // log warning or throw error or do nothing
   return;
}

new URL(...)
```

Alternatively, we can just let the code crash and have someone report it. Or maybe middle-ground, considering each global on a case-by-case basis.

Note: in the case of react-native specifically, we can [polyfill](https://www.npmjs.com/package/react-native-url-polyfill) just that bundle, but I don't like that we're operating blind. If we want to invest there, we'd ideally start with a test suite that can catch errors when unsupported globals are introduced to `coveo.analytics.js`, and only polyfill when we can guarantee that a) there is an error, and b) that adding the polyfill solves the error.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "coveo.analytics",
-    "version": "2.25.3",
+    "version": "2.26.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "coveo.analytics",
-            "version": "2.25.3",
+            "version": "2.26.0",
             "license": "MIT",
             "dependencies": {
                 "cross-fetch": "^3.1.5",
@@ -38,7 +38,6 @@
                 "prettier": "2.1.2",
                 "rimraf": "^3.0.2",
                 "rollup": "^2.50.6",
-                "rollup-plugin-polyfill-node": "^0.12.0",
                 "rollup-plugin-serve": "^1.0.1",
                 "rollup-plugin-terser": "^7.0.2",
                 "rollup-plugin-typescript2": "^0.27.0",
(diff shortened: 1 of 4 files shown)


rphilippen (approved):
I didn't know URL is available as a global in NodeJS? Or is it something that rollup detects for the build, injecting the module import for NodeJS? Regardless, changes look good.

### Aside 2
I have some objections to start using "fallback to `null`" accessor methods for globals, since you're basically just pushing the problem down a level. Now all code using the global has to deal with the situation that it may not be there (e.g. it may be `null`). This is fine for when the code can opt-in into a feature. (For example: if crypto is available, use secure random; otherwise, fall back to `Math.rand`). But if the dependency on the global is a hard requirement, I think we should focus on catching these issues using tests instead. Otherwise, all code that wishes to use the global has to take some action when the global is `null`, or we'll just get vague "null pointer exceptions" at runtime.

samisayegh:
> This is fine for when the code can opt-in into a feature.

Sounds good and agreed. The code would ideally not crash in a certain environment if it's possible to a) recover by using an alternative to the global, or b) if the pathway is not relevant for that environment (e.g. url processing is not relevant in the context or a mobile app built with react-native?).

Still there could be a nice benefit to centralize global access, retrieving them via functions: if different files rely on `URL`, we can look at one place and see what the assumption is, i.e. do we have a fallback in place, do we return null and let the caller deal with it, do we return the global always letting the code crash.

What do you think?

fvanham-coveo:
I'm closing this, since this is removing code that only was added in [UA-7368](https://github.com/coveo/coveo.analytics.js/pull/398). I've integrated these changes with the original PR.

PR #402, opened 2023-03-20
[UA-7587] React dependency bug

fvanham-coveo:
A dependency that should've been included in `react-native.es.js` **was not** (`react-native-get-random-values`), which is a polyfill for the browser's crypto.getRandomValues() which we're pulling through the common uuid dependency. See [here](https://github.com/uuidjs/uuid#react-native--expo) for more info.

Simultaneously, a dependency that should not have been there **was** included in `react-native.es.js`, namely `cross-fetch`. This happened because (the mistakenly added) `only` specification in rollup, part of deprecated `rollup-plugin-node-resolve`. The replacement library `@rollup/plugin-node-resolve` does not have this option, but it's silently ignored. In this PR, it's replaced by an `external` declaration. A new (external) dependency on `react-native` has been added since it's a dependency of the polyfill and assumed to be there.

Ultimately, we should have a react native test for the react-native integration to avoid these issues popping up in the future. It was non-trivial to add due to the large amounts of plugins and configs needed to run react native and node/ts side-by-side in jest. I diffed the generated file with the older known-good versions to verify the only additions are uuid and logic related.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -56,7 +56,7 @@ const browserUMD = {
     ],
     plugins: [
         browserFetch(),
-        nodeResolve({preferBuiltins: true, only: ['uuid']}),
+        nodeResolve({preferBuiltins: true, browser: true}),
         versionReplace(),
         tsPlugin(),
         process.env.SERVE
@@ -79,7 +79,7 @@ const nodeCJS = {
         format: 'cjs',
     },
     plugins: [
-        nodeResolve({mainFields: ['main'], preferBuiltins: true, only: ['uuid']}),
+        nodeResolve({mainFields: ['main'], preferBuiltins: true}),
         versionReplace(),
         commonjs(),
         tsPlugin(),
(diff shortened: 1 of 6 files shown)


aperron-coveo (approved):

lbourdages (approved):
image

rollup.config.js
```diff
@@ -105,13 +105,14 @@ const browserESM = {
 };
 
 const libRN = {
+    external: ['react-native', 'cross-fetch'],
```
samisayegh, line 108: What does the `react-native` entry do? I didn't spot an import on a dependency with that name so wasn't sure, hmm.

For `cross-fetch`, the dependency ships with polyfills for react-native (under `node_modules/cross-fetch/dist/react-native-ponyfill.js`). Wondering if we configure rollup to resolve and bundle that file?

samisayegh (approved):

rollup.config.js
fvanham-coveo: The react native entry is there because the UUID library needs a random generator, and it tries to [polyfill via the native functions](https://github.com/LinusU/react-native-get-random-values/blob/master/index.js) first. I think it's safe to say we can assume this library to be available externally.

Historically, coveo.analytics.js react native library has never had a fetch implementation bundled and we've had no complaints, so I'm not sure we should add it.

rphilippen (approved):

PR #407, opened 2023-04-13
fix: UA-7689 Convert product position and quantity to number if possible

aperron-coveo:
While this is technically not a bug, it is much easier for us to do this rather than asking every customer to send numbers instead of strings in those fields.

This simply checks if product quantity and position are of type "number" and if not we try to convert it. If the result is not a number, we send the original string to the collect endpoint and get the appropriate error from the backend.

- [x] Add tests

### DIFF
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -13,5 +13,6 @@ export function isObject(o: any): boolean {
  * @returns The input, possibly coerced to a number.
  */
 export function coerceToNumber(input: any): any {
-    return input != '' && !isNaN(input) ? +input : input;
+    // @ts-ignore
+    return typeof input === 'string' && input != '' && !isNaN(input) ? +input : input;
 }
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -13,11 +13,5 @@ export function isObject(o: any): boolean {
  * @returns The input, possibly coerced to a number.
  */
 export function coerceToNumber(input: any): any {
-    if (typeof input === 'string' && /^-?\d*\.?\d+$/.test(input)) {
-        const value = +input;
-        if (!Number.isNaN(value)) {
-            return value;
-        }
-    }
-    return input;
+    return input != '' && !isNaN(input) ? +input : input;
 }
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -13,6 +13,5 @@ export function isObject(o: any): boolean {
  * @returns The input, possibly coerced to a number.
  */
 export function coerceToNumber(input: any): any {
-    // @ts-ignore
-    return typeof input === 'string' && input != '' && !isNaN(input) ? +input : input;
+    return typeof input === 'string' && input != '' && !Number.isNaN(+input) ? +input : input;
 }
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -4,3 +4,20 @@ export const keysOf = Object.keys as <T>(o: T) => Extract<keyof T, string>[];
 export function isObject(o: any): boolean {
     return o !== null && typeof o === 'object' && !Array.isArray(o);
 }
+
+/**
+ * Attempt to coerce strings to number values following the "remove any prefix" logic.
+ * Any trailing non-number characters will fail parsing, in which case the input string is returned.
+ *
+ * @param input The input to possibly coerce to a number, if it is a string and parses to a number.
+ * @returns The input, possibly coerced to a number.
+ */
+export function coerceToNumber(input: any): any {
+    if (typeof input === 'string' && input.match(/^-?\d*\.?\d+$/)) {
+        const value = +input;
+        if (!Number.isNaN(value)) {
+            return value;
+        }
+    }
+    return input;
+}
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -6,14 +6,14 @@ export function isObject(o: any): boolean {
 }
 
 /**
- * Attempt to coerce strings to number values following the "remove any prefix" logic.
- * Any trailing non-number characters will fail parsing, in which case the input string is returned.
+ * Attempt to coerce strings to number values.
+ * Any non-number characters will fail parsing, in which case the input string is returned.
  *
  * @param input The input to possibly coerce to a number, if it is a string and parses to a number.
  * @returns The input, possibly coerced to a number.
  */
 export function coerceToNumber(input: any): any {
-    if (typeof input === 'string' && input.match(/^-?\d*\.?\d+$/)) {
+    if (typeof input === 'string' && /^-?\d*\.?\d+$/.test(input)) {
         const value = +input;
         if (!Number.isNaN(value)) {
             return value;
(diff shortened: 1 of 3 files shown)


lbourdages (commented):
Please add UTs 😄

src/plugins/ec.ts
```diff
@@ -160,11 +160,23 @@ export class ECPlugin extends BasePlugin {
...
                     ...newPayload,
-                    ...convertProductToMeasurementProtocol(product, index),
+                    ...convertProductToMeasurementProtocol(this.convertNumberTypes(product), index),
                 };
             }, {});
     }
 
+    private convertNumberTypes(product: Product) {
+        let updatedProduct = {...product};
+        if (Object.keys(product).indexOf('quantity') != -1 && typeof product.quantity != 'number') {
```
aperron-coveo, line 170: I'm not sure if it would be better here to do this instead
`Object.keys(product).indexOf('quantity') != -1 && typeof product.quantity === 'string'`
and then `// @ts-ignore` instead of casting as a string.

src/plugins/ec.ts
```diff
@@ -160,11 +160,23 @@ export class ECPlugin extends BasePlugin {
...
+            updatedProduct = {...updatedProduct, quantity: parseInt(String(product.quantity)) || product.quantity};
+        }
+        if (Object.keys(product).indexOf('position') != -1 && typeof product.position != 'number') {
+            updatedProduct = {...updatedProduct, position: parseInt(String(product.position)) || product.position};
+        }
+
+        return updatedProduct;
```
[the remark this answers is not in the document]
aperron-coveo: I'm a bit confused about what the regex does in this. It seems to me that we can potentially replace user input with something else. For example, if the user passes 'abc', the regex replaces this by '' and then the `+` operator converts that to 0. Maybe we could just skip the conversion if it matches `[^\.\-0-9]`?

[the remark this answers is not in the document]
aperron-coveo:
```
private coerceToNumber(input: any): any {
    if (typeof input === 'string' && !input.match(/[^\.\-0-9]/) && input != '') {
        const value = +input;
        if (!isNaN(value)) {
            return value;
        }
    }
    return input;
}
```

src/plugins/ec.ts
fvanham-coveo, line 170: Don't iterate all the keys. Just snipe the keys you are looking for (`quantity` and `position`):

src/plugins/ec.ts
```diff
@@ -160,11 +160,23 @@ export class ECPlugin extends BasePlugin {
             .reduce((newPayload, product, index) => {
                 return {
                     ...newPayload,
-                    ...convertProductToMeasurementProtocol(product, index),
+                    ...convertProductToMeasurementProtocol(this.convertNumberTypes(product), index),
                 };
             }, {});
     }
 
+    private convertNumberTypes(product: Product) {
```
fvanham-coveo, line 168: missing return type?

src/plugins/ec.ts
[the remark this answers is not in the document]
fvanham-coveo: The prefix argument is valid for currency type numbers only. We accept string values which include non-numeric prefix according to doc here: https://docs.coveo.com/en/l29e0540/coveo-for-commerce/commerce-coveo-ua-api-reference#currency
This does not apply to regular numbers like quantity and position. I would not do the regexes. If someone sends 💩 string they get 💩 string in the validation :)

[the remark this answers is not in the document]
fvanham-coveo: A coerce function is a good idea, there's already a `utils.ts`. I'd make it so that a coerce on `undefined` just returns `undefined`, then you can blindly call `product.position = coerceToNumber(product.position)` in your main proc. If the key does not exist it won't be added.

fvanham-coveo (changes requested):

src/plugins/ec.ts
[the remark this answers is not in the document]
aperron-coveo: With the way I modified it, the regex checks for a char that is not a number or a `-` and if it matches anything, we just return the string. And for the "returning undefined" part, I'm pretty sure it would already work as is, so I'll remove the checks (`'quantity' in updatedProduct`) and move the coerce method to utils.ts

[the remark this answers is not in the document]
rphilippen: Frank, just returning undefined and always setting it has the unfortunate side effect of **adding** a property, even if it wasn't specified. I don't think we should do that.

> If the key does not exist it won't be added.
That's not how JavaScript works:
```javascript
'product' in {product: undefined} === true
```

rphilippen (approved):

src/client/utils.ts
```diff
@@ -4,3 +4,20 @@ export const keysOf = Object.keys as <T>(o: T) => Extract<keyof T, string>[];
...
+
+/**
+ * Attempt to coerce strings to number values following the "remove any prefix" logic.
+ * Any trailing non-number characters will fail parsing, in which case the input string is returned.
+ *
+ * @param input The input to possibly coerce to a number, if it is a string and parses to a number.
+ * @returns The input, possibly coerced to a number.
+ */
+export function coerceToNumber(input: any): any {
+    if (typeof input === 'string' && input.match(/^-?\d*\.?\d+$/)) {
```
fvanham-coveo: I don't think we need the (expensive) regex check here? Regardless, you're missing scientific notation numbers.

src/plugins/ec.spec.ts
```diff
@@ -550,4 +530,36 @@ describe('EC plugin', () => {
         afterHook(eventType, result);
         return result;
     };
+
+    const testValidNumberConversion = (source: keyof Product, target: string) => {
+        const validValues = ['13', '0', '.0', '-10', '5.0', '-1.1'];
```
fvanham-coveo, line 535: This is preference, but I don't mind it if tests are not DRY. I now have to go to the test main func, find the called method, see what it does, wonder why you are deleting targets at the end. I came to the conclusion that that is because you are passing negative values into `position` due to reusing your test array and there's logic inside the library which removes that value, but only for position.

IMO I liked your first test setup way better. Could've done it with multiple products if you wanted a broader set of numbers.

src/client/utils.ts
aperron-coveo: Do you have any idea other than with a regex to check if the input can be parsed?

src/client/utils.ts
```diff
@@ -4,3 +4,14 @@ export const keysOf = Object.keys as <T>(o: T) => Extract<keyof T, string>[];
...
+/**
+ * Attempt to coerce strings to number values.
+ * Any non-number characters will fail parsing, in which case the input string is returned.
+ *
+ * @param input The input to possibly coerce to a number, if it is a string and parses to a number.
+ * @returns The input, possibly coerced to a number.
+ */
+export function coerceToNumber(input: any): any {
+    return input != '' && !isNaN(input) ? +input : input;
+}
```
aperron-coveo, line 17: @rphilippen @fvanham-coveo this allows for numbers like `1.234e5` and `5.` to be converted. Its also much simpler than the regex

fvanham-coveo (approved):

rphilippen (approved):

PR #408, opened 2023-04-14
[UA-7720] update coveoua dependencies

fvanham-coveo:
This change updates the dependencies of coveoua to more recent versions. Many dev dependencies were out of support, had critical vulnerabilities or did not support more modern versions of key tooling like typescript.

Notes:
- Fixed an incorrect test (was testing impression, should've been testing product)
-  I added event types to library so that direct library users have typing for them.
- Small improvement in error handling related to this [maintenance case](https://coveord.atlassian.net/browse/UA-7700).
- All tests pass, I manually verified the new version against a couple of customer integrations using Chrome only. The newly generated js is 1kb bigger (97K).


jmprovencher (approved):
I would be interesting in knowning how you test that on some customers live on their website

lbourdages (approved):

rphilippen (approved):

fvanham-coveo:
> I would be interesting in knowning how you test that on some customers live on their website

See `readme.md`. It's using a Chrome plugin called redirector, which redirects (ahem) any http request to `static.cloud.coveo.com/coveo.analytics.js/` to my localhost instead.

PR #412, opened 2023-04-25
fix(SFINT-4992): fix a crash in HistoryStore

erocheleau:
[SFINT-4992](https://coveord.atlassian.net/browse/SFINT-4992)

For whatever reason, the value in `__coveo.analytics.history` in the localStorage when is valid JSON but NOT an array, makes the HistoryStore crash because it tries to call `.sort` and `.map` on something that isn't enforced to be an array.

I added safeguards against this, and a test to make sure it was covered.

[SFINT-4992]: https://coveord.atlassian.net/browse/SFINT-4992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/history.ts
+++ b/src/history.ts
@@ -103,7 +103,7 @@ export class HistoryStore {
 
     getMostRecentElement(): HistoryElement | null {
         let currentHistory = this.getHistoryWithInternalTime();
-        if (currentHistory != null) {
+        if (Array.isArray(currentHistory)) {
             const sorted = currentHistory.sort((first: HistoryElement, second: HistoryElement) => {
                 // Internal time might not be set for all history element (on upgrade).
                 // Ensure to return the most recent element for which we have a value for internalTime.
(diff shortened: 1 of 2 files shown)


lbourdages (approved):
Can this happen in normal operation, or just when someone messes with their local storage?

src/history.ts
```diff
@@ -103,7 +103,7 @@ export class HistoryStore {
 
     getMostRecentElement(): HistoryElement | null {
         let currentHistory = this.getHistoryWithInternalTime();
-        if (currentHistory != null) {
+        if (Array.isArray(currentHistory)) {
```
tedre191, line 106: As this shouldn't happen unless something else is broken, and that is also mean that part of the feature will not work, we should add a warning in the console to let support, or the next person in maintenance, that will have to debug this in the future.

tedre191 (changes requested):

src/history.ts
erocheleau: I don't exactly know what the feature does, this is a HistoryClient used in the JSUI for example, so the JSUI handles what happens when the historyclient returns an empty array basically, and not the coveoua library

erocheleau:
> Can this happen in normal operation, or just when someone messes with their local storage?

We've had a support case because of some weird interaction inside Salesforce and their Locker Service, two localStorage keys had a collision basically. So some other data ended up in this localStorage key, which was valid JSON, but not an array.

And the crash from the lib was crashing the whole search page basically.

Overall I think it's best to make sure something is actually an array before trying to use Array functions on it :)

tedre191 (approved):

PR #413, opened 2023-05-01
Adding support for reverse-proxies in analytics.js library.

jmprovencher:
image (4)

We're working on adding documentation on how to serve analytics through a reverse-proxy for some of our customers. One problem with that approach is that when a customer would set an endpoint like this,

`coveoua('init', "xxd3323beb-702b-49bf-b563-698604eef887", "https://dwt17j4ahkcnx.cloudfront.net/rest/ua", );`

the effective endpoint would become "https://dwt17j4ahkcnx.cloudfront.net/rest/ua/rest"

This feature aims at fixing this behaviour, but also tries to improve the default logic when passing endpoints instead on relying on a condition to check if the url contains .cloud.coveo.com

It will effectively do the following.

1. Remove any trailing slash
2. If it's a custom endpoint, it will take the endpoint as is, and add the api version to it
3. if the endpoint has "/rest/ua" in it, it will effectively add the api version number to it, and return it
4. if the endpoint does not have /rest/ua in it, and its not a custom endpoint, it will add it to it.

This should support all the following endpoints that could be currently passed by customers.

- https://usageanalytics.com
- https://usageanalytics.com/
- https://usageanalytics.com/rest/ua
- https://analytics.cloud.coveo.com/rest/ua
- https://analytics.cloud.coveo.com/
- https://analytics.cloud.coveo.com

IT WILL NOT WORK if somebody would pass 

https://usageanalytics.com/rest but it would already not be working so that's ok.

This will allow customer to customize and endpoint like coveo.com/analytics that would be forwarded to analytics.cloud.coveo.com/rest/ua if they want to implement a reverse-proxy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,10 @@ After the library has loaded sucessfully, you can interact with coveoua through 
 -   `coveoua('init', <COVEO_API_KEY>, <ENDPOINT>)`: Initializes the library with the given api key and endpoint. The following parameters are accepted
     -   COVEO_API_KEY (mandatory): A valid api key.
     -   ENDPOINT (optional): A string specifying the desired analytics endpoint. The default value is https://analytics.cloud.coveo.com/rest/ua. In case your organization is HIPAA enabled, you should override with https://analyticshipaa.cloud.coveo.com/rest/ua.
--   `coveoua('init', <COVEO_API_KEY>, {endpoint: <ENDPOINT>, plugins: <PLUGINS>})`: Initializes the library with the given api key, endpoint and plugins. The following parameters are accepted
+-   `coveoua('init', <COVEO_API_KEY>, {endpoint: <ENDPOINT>, isCustomEndpoint: <IS_CUSTOM_ENDPOINT>, plugins: <PLUGINS>})`: Initializes the library with the given api key, endpoint, isCustomEndpoint option and plugins. The following parameters are accepted
     -   COVEO_API_KEY (mandatory): A valid api key.
     -   ENDPOINT (optional): An object string specifying the desired analytics endpoint. The default value is https://analytics.cloud.coveo.com/rest/ua. In case your organization is HIPAA enabled, you should override with https://analyticshipaa.cloud.coveo.com/rest/ua.
+    -   IS_CUSTOM_ENDPOINT (optional): An boolean specifying if the desired analytics endpoint is custom. This means the library will not try to add any prefix logic like /rest/ua at the end of the provided endpoint. This can be useful when implementing analytics through a reverse-proxy solution like cloudfront.
     -   PLUGINS (optional): An array of known plugin names. See [plugins](#plugins) for more information.
 -   `coveoua('set', <NAME>, <VALUE>)`: Attempts to inject an attribute with given name and value on every logged event, overriding any existing value. Some payloads may reject attributes they do not support.
 -   `coveoua('set', <OBJECT>)`: Attempts to inject all attributes and values of the given object on every logged event, overriding any existing value. Some payloads may reject attributes they do not support.
--- a/src/coveoua/simpleanalytics.ts
+++ b/src/coveoua/simpleanalytics.ts
@@ -9,6 +9,7 @@ export type AvailableActions = keyof CoveoUA;
 
 export interface CoveoUAOptions {
     endpoint?: string;
+    isCustomEndpoint?: boolean;
     plugins?: string[];
 }
 
@@ -31,6 +32,7 @@ export class CoveoUA {
             this.client = new CoveoAnalyticsClient({
                 token: token,
                 endpoint: this.getEndpoint(optionsOrEndpoint),
+                isCustomEndpoint: this.getIsCustomEndpoint(optionsOrEndpoint),
             });
         } else if (this.isAnalyticsClient(token)) {
             this.client = token;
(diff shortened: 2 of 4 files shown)


lbourdages:
> IT WILL NOT WORK if somebody would pass
> 
> https://usageanalytics.com/rest but it would already not be working so that's ok.

Why don't we just support it?

/ua is not required anymore. To me the check could be as simple as `if(!url.contains("/rest")) -> add /rest/ua`.

jmprovencher:
> 

@lbourdages well it's required if you go through platform.cloud.coveo.com

lbourdages:
> > 
> 
> @lbourdages well it's required if you go through platform.cloud.coveo.com
> 
> if you add /rest to platform.cloud.coveo.com, it will not work, which would be a sneaky corner-case that I'd rather not handle improperly.

It's not???

https://platform.cloud.coveo.com/rest/v15/analytics/status

lbourdages:
Like, I know it works, I'm the one who made it work 😛 There is only one path now, it's all going through the proxy, it all works in the same way

jmprovencher:
> Like, I know it works, I'm the one who made it work 😛 There is only one path now, it's all going through the proxy, it all works in the same way

you're right, but well I wanted to replicate what seemed to be the "official" path (from the swagger it's /rest/ua )

lbourdages:
> > Like, I know it works, I'm the one who made it work 😛 There is only one path now, it's all going through the proxy, it all works in the same way
> 
> you're right, but well I wanted to replicate what seemed to be the "official" path (from the swagger it's /rest/ua ).
> 
> I'll try to change my solution to do what you propose.

I think it's better to use /rest/ua but ideally we should try to support /rest if we can.

For reference: https://github.com/coveo/proxy/blob/develop/proxy/src/main/resources/application.yml#LL664C1-L682C1

jmprovencher:
> > > Like, I know it works, I'm the one who made it work 😛 There is only one path now, it's all going through the proxy, it all works in the same way
> > 
> > 
> > you're right, but well I wanted to replicate what seemed to be the "official" path (from the swagger it's /rest/ua ).
> > I'll try to change my solution to do what you propose.
> 
> I think it's better to use /rest/ua but ideally we should try to support /rest if we can.
> 
> For reference: https://github.com/coveo/proxy/blob/develop/proxy/src/main/resources/application.yml#LL664C1-L682C1

yeah, I can to replace any endpoint that do not contain "/rest" and that don't have the flag "isCustomEndpoint" by an endpoint with "/rest/ua" at the end.

src/coveoua/simpleanalytics.ts
```diff
@@ -72,9 +74,18 @@ export class CoveoUA {
         return Endpoints.default;
     }
 
+    private getIsCustomEndpoint(optionsOrEndpoint: string | CoveoUAOptions) {
```
lbourdages, line 77: today I discovered the existence of the pipe operator in typescript and I died a little inside

lbourdages (approved):
Looks good but I find the init code with the or types really sketchy... however it was mostly like this before your PR. Worth a cleanup eventually IMO

Enteris (approved):

README.md
```diff
@@ -47,9 +47,10 @@ After the library has loaded sucessfully, you can interact with coveoua through
...
--   `coveoua('init', <COVEO_API_KEY>, {endpoint: <ENDPOINT>, plugins: <PLUGINS>})`: Initializes the library with the given api key, endpoint and plugins. The following parameters are accepted
+-   `coveoua('init', <COVEO_API_KEY>, {endpoint: <ENDPOINT>, isCustomEndpoint: <IS_CUSTOM_ENDPOINT>, plugins: <PLUGINS>})`: Initializes the library with the given api key, endpoint, isCustomEndpoint option and plugins. The following parameters are accepted
     -   COVEO_API_KEY (mandatory): A valid api key.
     -   ENDPOINT (optional): An object string specifying the desired analytics endpoint. The default value is https://analytics.cloud.coveo.com/rest/ua. In case your organization is HIPAA enabled, you should override with https://analyticshipaa.cloud.coveo.com/rest/ua.
+    -   IS_CUSTOM_ENDPOINT (optional): An boolean specifying if the desired analytics endpoint is custom. This means the library will not try to add any prefix logic like /rest/ua at the end of the provided endpoint. This can be useful when implementing analytics through a reverse-proxy solution like cloudfront.
```
samisayegh, line 53: Could we avoid the `isCustomEndpoint` option by checking if the url matches a list of urls we support?
It would one less thing to specify/document, but also captures the list of urls we do support in code.

jmprovencher: by looking at the previous logic, you can see that doing something like that poorly ages IMO because if there's a new url at some point then it does not work as expected. I'd rather have a declarative argument for that.

samisayegh: It's not ideal though that one could pass a custom url and forget to specify `IS_CUSTOM_ENDPOINT`, or pass a Coveo url, but set `IS_CUSTOM_ENDPOINT` to `true`. It's declarative, but an incorrect combination of parameters.

If we know the full list of Coveo urls (declarative), or if we check for `.coveo.com` in the url (imperative), it feels like could avoid implementers configuring invalid combinations. There are tradeoffs here too though (company changes it's name, or we accept events on domains other than `.coveo.com`). Just some thoughts, I leave it to you,

samisayegh (approved):

README.md
jmprovencher: This isCustomEndpoint is meant for somebody implementing a reverse-proxy. If somebody "forgets" to pass a parameter, his implementation will fail from the get go. I understand your point, however, when me and Richard actually were working on this as a "mock" customer, the fact that there was internal logic doing something different based on the url we pass, without being documented, and not stated in any way in the attributes we pass, made us lose a lot of time.

To be honest, I think the whole way URL are handles in this library are far from ideal. I think it's a lesson for the new library we're building. I think we should always specifiy the whole path and not add any suffix, this way there's a one size fits all method to passing an endpoint.

PR #414, opened 2023-05-03
Upgrade node to 14 (buster/bullseye)

fvanham-coveo:
The build is broken because (I think) the Debian distro (Stretch) that came with node-10 is longer in LTS support. Node is used to run the FT's on the backend. apt-get update fails as a result. This change upgrade the node version to 14, which should upgrade the Debian version to at least Buster.

### DIFF
--- a/cypress/Dockerfile
+++ b/cypress/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:14
 
 RUN apt-get update && \
     apt-get -y install xvfb gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \


cypress/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:10
+FROM node:14
```
rphilippen, line 1: Just curious: why not 18, to be a bit more current? It is the current Long Term Support version. Or even 16, to use a less-old-but-still-old version?

rphilippen (approved):

lbourdages (approved):

cypress/Dockerfile
fvanham-coveo: Mostly because I don't want to make too big a step at once and end up in outdated-dependecy-hell.

PR #415, opened 2023-05-03
[UA-7777] fix coveoua build

fvanham-coveo:
After upgrading to node14 the build continues, but Cypress now has a few missing dependencies. The new ones added comes from this list:
https://docs.cypress.io/guides/continuous-integration/introduction#Dependencies
I did not remove older ones as I'm not sure why they were added in the first place.

### DIFF
--- a/cypress/Dockerfile
+++ b/cypress/Dockerfile
@@ -1,7 +1,8 @@
 FROM node:14
 
 RUN apt-get update && \
-    apt-get -y install xvfb gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \
+    apt-get -y install xvfb libgtk2.0-0 libgbm-dev libnotify-dev libxss1 xauth \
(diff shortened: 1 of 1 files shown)


AdamMartineau-Coveo (approved):

cypress/Dockerfile
```diff
@@ -1,7 +1,8 @@
 FROM node:14
 
 RUN apt-get update && \
-    apt-get -y install xvfb gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \
+    apt-get -y install xvfb libgtk2.0-0 libgbm-dev libnotify-dev libxss1 xauth \
+      gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \
```
mpayne-coveo, line 5: Just curious, how big does this make the Dockerfile? (not that this is a huge issue since this is just Cypress)

fvanham-coveo: I have no idea :)

PR #428, opened 2023-06-27
support trackingId

lucylu-coveo:
### [UA-7856](https://coveord.atlassian.net/browse/UA-7856)
1.  add trackingId in the payload when sending events
2. `trackingId` in the payload is set to `trackingId` or `context_website`. When `trackingId` is not set explicitly AND `context_website` is present, `trackingId` will be set with `context_website` 

[UA-7856]: https://coveord.atlassian.net/browse/UA-7856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/analytics.spec.ts
+++ b/src/client/analytics.spec.ts
@@ -439,6 +439,29 @@ describe('Analytics', () => {
         });
     });
 
+    describe('with context_website is set', () => {
+        const contextWebsite = 'yourbestfriend.com';
+        const trackingId = 'yourfavaritefood.ca ';
+        beforeEach(() => {
+            client = new CoveoAnalyticsClient({
+                token: 'xtoken',
+                endpoint: anEndpoint,
+                version: A_VERSION,
+            });
+            mockFetchRequestForEventType(EventType.view);
+        });
+        it('should set trackingId when trackingId is not specified', async () => {
+            await client.sendEvent(EventType.view, {custom: {context_website: contextWebsite}});
+            const [body] = getParsedBodyCalls();
+            expect(body.trackingId).toBe(contextWebsite);
+        });
+        it('should not set trackingId when trackingId is specified', async () => {
+            await client.sendEvent(EventType.view, {trackingId: trackingId, custom: {context_website: contextWebsite}});
+            const [body] = getParsedBodyCalls();
+            expect(body.trackingId).toBe(trackingId);
+        });
+    });
+
     it('should support clearing cookies for visitorId and historyStore', () => {
         const visitorId = 'foo';
         const history = {name: 'foo', time: '123', value: 'bar'};
--- a/src/client/analytics.spec.ts
+++ b/src/client/analytics.spec.ts
@@ -450,11 +450,13 @@ describe('Analytics', () => {
             });
             mockFetchRequestForEventType(EventType.view);
         });
+
         it('should set trackingId when trackingId is not specified', async () => {
             await client.sendEvent(EventType.view, {custom: {context_website: contextWebsite}});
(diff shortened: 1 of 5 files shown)


src/client/analytics.ts
```diff
@@ -623,6 +624,16 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         }
     }
 
+    private setTrackingId(payload: IRequestPayload): IRequestPayload {
```
lbourdages, line 627: setTrackingIdFromContextWebsiteIfTrackingIdNotPresent? 

A bit verbose but it would be more explicit 😄

lbourdages (approved):

jmprovencher (approved):

src/client/analytics.ts
```diff
@@ -623,6 +625,16 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider
         }
     }
 
+    private setTrackingIdFromContextWebsiteIfTrackingIdNotPresent(payload: IRequestPayload): IRequestPayload {
+        const {trackingId, custom, ...rest} = payload;
+        if (!trackingId && custom && isObject(custom) && 'context_website' in custom) {
+            rest['trackingId'] = custom.context_website;
```
fvanham-coveo, line 631: We have 261 orgs in us/east which us `siteName` instead of `context_website` because of a doc mess-up. Is this "officially" supported? Currently no. Do we have to deal with it? I think so, unfortunately.

src/client/analytics.spec.ts
```diff
@@ -439,6 +439,31 @@ describe('Analytics', () => {
...
+            client = new CoveoAnalyticsClient({
+                token: 'xtoken',
+                endpoint: anEndpoint,
+                version: A_VERSION,
+            });
+            mockFetchRequestForEventType(EventType.view);
+        });
+
+        it('should set trackingId when trackingId is not specified', async () => {
+            await client.sendEvent(EventType.view, {custom: {context_website: contextWebsite}});
```
fvanham-coveo, line 455: This is incorrect. `custom` param should read `customData` I think.

fvanham-coveo:
Recommend we revert this change, fix `siteName` and the `custom` key and then resubmit for review.

src/client/analytics.spec.ts
lucylu-coveo: `context_website` only ends up in `customData` after [mapCustomParametersToCustomData](https://github.com/coveo/coveo.analytics.js/blob/c19bc6a2cfb6cf5ea4c43707faa8b07e10429ec3/src/client/analytics.ts#L572) is invoked. (already discussed with Frank, just put it here for future record)

lucylu-coveo: But the conversion happens after function call `setTrackingIdFromContextWebsiteIfTrackingIdNotPresent` inside [resolvePayloadForParameters](https://github.com/coveo/coveo.analytics.js/blob/c19bc6a2cfb6cf5ea4c43707faa8b07e10429ec3/src/client/analytics.ts#L344). My understanding is that`{custom: {context_website: contextWebsite}` is the original payload, and it will eventually be converted to `customData: {context_website: contextWebsite}` (returned by resolvePayloadForParameters). I have tested locally and verified that the eventual payload (forwarded to usageanalytics api ) is correct for collect and non-collect events. I'm still learning Javascript/Typescript 😛 and happy to look at it together

PR #431, opened 2023-07-06
set trackingid from siteName and properties in customData

lucylu-coveo:
1. set trackingId with `siteName`
2. support extracting `siteName` and `context_website` from customData object

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -570,6 +570,38 @@ describe('ec events', () => {
         });
     });
 
+    it('should be able to set trackingId with siteName', async () => {
+        const website = 'tracksite';
+        await coveoua('set', 'custom', {siteName: website});
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            trackingId: website,
+            sitename: website,
+        });
+    });
+
+    it('siteName does not overwrite trackingId', async () => {
+        const website = 'tracksite';
+        const trackingId = 'trackingid';
+        await coveoua('set', 'custom', {siteName: website});
+        await coveoua('set', 'trackingId', trackingId);
+        await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            trackingId: trackingId,
+            sitename: website,
+        });
+    });
+
     describe('with auto-detection of userId', () => {
         describe('with API key', () => {
             beforeEach(() => {
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -538,10 +538,25 @@ describe('ec events', () => {
         });
     });
 
-    it('should be able to set trackingId with context_website', async () => {
+    it('should be able to set trackingId with context_website for collect event', async () => {
         const contextWebsite = 'tracksite';
         await coveoua('set', 'custom', {context_website: contextWebsite});
         await coveoua('send', 'pageview');
+
+        const [event] = getParsedBody();
+
+        expect(event).toEqual({
+            ...defaultContextValues,
+            t: 'pageview',
+            trackingId: contextWebsite,
+            context_website: contextWebsite,
+        });
+    });
+
+    it('should be able to set trackingId with context_website for non-collect event', async () => {
+        const contextWebsite = 'tracksite';
+        await coveoua('set', 'custom', {context_website: contextWebsite});
+        await coveoua('send', 'view');
 
         const [event] = getParsedBody();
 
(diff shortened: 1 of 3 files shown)


functional/ec-events.spec.ts
```diff
@@ -570,6 +570,38 @@ describe('ec events', () => {
         });
     });
 
+    it('should be able to set trackingId with siteName', async () => {
```
fvanham-coveo, line 573: These tests look good. I'd add a test that calls `await coveoua('send', 'view')` and verifies the element ends up in `customData` instead of the root, for both cases.

note: view is a UA event, pageview is a collect event.

src/client/analytics.spec.ts
```diff
@@ -439,7 +439,7 @@ describe('Analytics', () => {
         });
     });
 
-    describe('with context_website is set', () => {
+    describe('with context_website is set in custom', () => {
```
fvanham-coveo, line 442: This test still tests the incorrect condition: `custom` is not an officially supported key on `EventType.view`, that should read `customData`, but you've added these tests below I see. Even if this behavior currently works, it's not something we should encourage or ensure keeps working through a test.

src/client/analytics.spec.ts
```diff
@@ -457,13 +457,91 @@ describe('Analytics', () => {
...
+        it('should not overwrite trackingId when trackingId is specified', async () => {
             await client.sendEvent(EventType.view, {trackingId: trackingId, custom: {context_website: contextWebsite}});
             const [body] = getParsedBodyCalls();
             expect(body.trackingId).toBe(trackingId);
         });
     });
 
+    describe('with context_website is set in customData', () => {
+        const contextWebsite = 'yourbestfriend.com';
+        const trackingId = 'yourfavaritefood.ca ';
```
fvanham-coveo, line 469: favorite

src/client/analytics.spec.ts
```diff
@@ -457,13 +457,91 @@ describe('Analytics', () => {
...
+            await client.sendEvent(EventType.view, {
+                trackingId: trackingId,
+                customData: {context_website: contextWebsite},
+            });
+            const [body] = getParsedBodyCalls();
+            expect(body.trackingId).toBe(trackingId);
+        });
+    });
+
+    describe('with siteName is set in custom', () => {
```
fvanham-coveo, line 495: Same argument as above: `custom` is not a supported key on the event payload

src/client/analytics.spec.ts
```diff
@@ -457,13 +457,91 @@ describe('Analytics', () => {
...
+        });
+
+        it('should set trackingId when trackingId is not specified', async () => {
+            await client.sendEvent(EventType.view, {customData: {context_website: contextWebsite}});
+            const [body] = getParsedBodyCalls();
+            expect(body.trackingId).toBe(contextWebsite);
+        });
+
+        it('should not overwrite trackingId when trackingId is specified', async () => {
+            await client.sendEvent(EventType.view, {
```
fvanham-coveo, line 486: I would also test this with a collect event (e.g. `EventType.pageview`) where the context_website is set on the root. If trackingId is not set, it should set it. If trackingId is also set it should not overwrite it.

fvanham-coveo (changes requested):

src/client/analytics.spec.ts
lucylu-coveo: this suite is testing this branch in setTrackingIdIfTrackingIdNotPresent:
```
        if (rest.hasOwnProperty('custom') && isObject(rest.custom)) {
            if (rest.custom.hasOwnProperty('context_website') || rest.custom.hasOwnProperty('siteName')) {
                rest['trackingId'] = rest.custom.context_website || rest.custom.siteName;
            }
        }
```

src/client/analytics.spec.ts
lucylu-coveo: I'm not sure why we need to test context_website set on the root level? :thinking `trackingId` is only set with `context_website` in either `custom` or `customData` object. I have added the test as you suggested in my latest [commit](https://github.com/coveo/coveo.analytics.js/pull/431/commits/1d862cbfad63f11580d481eb271fb7575f88818c). It fails  as I expect 😛

fvanham-coveo: Sorry, then the only way to set this for collect is by setting it as `coveoua('set', 'custom', {})`. My bad.

src/client/analytics.spec.ts
fvanham-coveo: Yeah, but that branch is an officially unsupported case. `custom` is not a recognized event attribute on any event. It's a side effect of the original implementers injecting the `custom` properties on the event itself, rather than separating it out. If we add it as a test, that internal side-effect becomes sanctioned external behavior.

fvanham-coveo (approved):

aperron-coveo (approved):

PR #433, opened 2023-08-07
added "modules" sub-package

btaillon:
https://coveord.atlassian.net/browse/KIT-2003

When importing `coveo.analytics` with Next.js, we get an error on the client side. This is because, on the client side, Next.js imports the "browser" field from the `package.json` file. In `coveo.analytics`, this field refers to the `coveoua` Javascript file, which doesn't have the same exports as the type definitions and globally defines `coveoua`.

Additionally, unlike the "main" export, the "module" export references the global `window` object.

In this PR, I did three things:
1. I renamed the entrypoints so they reflect their real use case
   * `library.es.js` became `browser.mjs`
   * `library.js` became `library.cjs`.
2. I ensured there's a corresponding ESM/CJS output for every output.
   * I added `library.mjs` and `browser.cjs`
3. I added a "modules` sub-package which exports the same modules as the type definitions across the board.
   * Can be imported using `import { someStuff } from 'coveo.analytics/modules'`.
   * I added this module instead of fixing the fields in the original `package.json` since doing so would've been a breaking change.

I'm not 100% certain if I want to merge this PR. There are ways to work around this issue in implementations, and exporting a new sub-package is pretty ugly. I'm curious what you guys think.

### DIFF
--- a/modules/package.json
+++ b/modules/package.json
@@ -1,8 +1,9 @@
 {
     "name": "modules",
     "private": true,
-    "main": "../dist/library.js",
-    "browser": "../dist/library.es.js",
+    "main": "../dist/library.cjs",
+    "module": "../dist/library.mjs",
+    "browser": "../dist/browser.mjs",
     "types": "./../dist/definitions/coveoua/library.d.ts",
     "files": [
         "../dist"
--- a/modules/package.json
+++ b/modules/package.json
@@ -4,7 +4,7 @@
     "main": "../dist/library.cjs",
     "module": "../dist/library.mjs",
     "browser": "../dist/browser.mjs",
-    "types": "./../dist/definitions/coveoua/library.d.ts",
+    "types": "../dist/definitions/coveoua/library.d.ts",
     "files": [
         "../dist"
     ]
--- a/modules/package.json
+++ b/modules/package.json
@@ -1,0 +1,10 @@
+{
+    "name": "modules",
+    "private": true,
+    "main": "../dist/library.js",
+    "browser": "../dist/library.es.js",
+    "types": "./../dist/definitions/coveoua/library.d.ts",
+    "files": [
+        "../dist"
+    ]
+}
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
         "plist": "3.0.5"
     },
     "files": [
+        "modules/",
         "dist/**/*.d.ts",
         "dist/**/*.js",
         "dist/**/*.js.map",
--- a/package.json
+++ b/package.json
@@ -2,8 +2,8 @@
     "name": "coveo.analytics",
     "version": "2.28.12",
     "description": "📈 Coveo analytics client (node and browser compatible) ",
-    "main": "dist/library.js",
-    "module": "dist/library.es.js",
+    "main": "dist/library.cjs",
+    "module": "dist/browser.mjs",
     "browser": "dist/coveoua.browser.js",
     "types": "dist/definitions/coveoua/library.d.ts",
     "scripts": {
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -2,8 +2,8 @@
     "name": "coveo.analytics",
     "version": "2.28.12",
     "description": "📈 Coveo analytics client (node and browser compatible) ",
-    "main": "dist/library.js",
-    "module": "dist/library.es.js",
+    "main": "dist/library.cjs",
+    "module": "dist/browser.mjs",
```
samisayegh, line 6: I'm questioning if the re-name is a step in the right direction. Seeing `browser` next to `mjs` feels a bit contradictory, but also I think the intent is to expose a file that will work in both the browser and node, not just one or the other?

On the other hand, it seems like the `libray.js` and `library.es.js` are generated using different entry points, so naming both files "library" does not seem accurate either?

modules/package.json
```diff
@@ -0,0 +1,11 @@
+{
+    "name": "modules",
+    "private": true,
+    "main": "../dist/library.cjs",
+    "module": "../dist/library.mjs",
+    "browser": "../dist/browser.mjs",
```
samisayegh, line 6: What if we leveraged the [exports](https://nodejs.org/api/packages.html#package-entry-points) property inside the existing `package.json` file instead of creating a new package.json? Could that be a path to solving the issue in a non-breaking way? It's hard to tell without testing, but it feels like it could work, hmm

Example config from [relay](https://github.com/coveo/relay/pull/7/files#diff-35ae7e0e77875d37217ddf591cc04926d64bee5fc3bc6e5fd47bf40f786247b5R10)

package.json
louis-bompart: > Seeing the browser next to mjs feels a bit contradictory
IMO, it's the contrary, it's a pleonasm, as modules, i.e. ESM, is the only "module" supported natively by the browser (without hoops and hiccups)

rollup.config.mjs
```diff
@@ -90,12 +100,15 @@ const nodeCJS = {
...
+ * @satisfies {RollupOptions}
+ */
+const browserModulesConfig = {
     input: './src/coveoua/headless.ts',
-    output: {
-        file: './dist/library.es.js',
-        format: 'es',
-    },
+    output: /** @type {const} */ (['cjs', 'es']).map((format) => ({
+        file: `./dist/browser.${format === 'cjs' ? 'cjs' : 'mjs'}`,
```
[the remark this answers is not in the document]
btaillon: > It's strange to see `.cjs` and `.mjs` in the config for a browser bundle because these are for nodejs.

I wasn't aware that those are for nodejs, especially `mjs`. Do you know a spec where it mentions whether they are supported? When I [test in codesandbox](https://8rmmct.csb.app/), it seems to work, and I see `.mjs` in the [common MIME types](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), but that's it.

> Why not just bundle as es and use a .js extension for the browser?

My choice was mostly just semantic. I knew that Node.js sometimes uses the extension to determine whether a file uses CommonJS or ESM, so it felt like an established standard and I defaulted to that choice.

I don't mind renaming it to something else like you said.

package.json
btaillon: > Seeing `browser` [...] I think the intent is to expose a file that will work in both the browser and node, not just one or the other?

Unfortunately, this bundle uses `browserFetch()` (see `rollup.config.js`), which directly references `window.fetch`.

rollup.config.mjs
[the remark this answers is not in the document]
btaillon: edit: for now, I'll keep it as `mjs` since it may help bundlers infer what's the type of export in this file. I'll remove the commonJS export though.

fvanham-coveo (approved):

fvanham-coveo:
I compared the output of the new build with the output of the old build, and they're identical (minus the version identifier). There's no risk in merging this package, but it's a bit weird to have a second package.json on the side. I don't know what we'll be breaking by modifying the original package.json though.

samisayegh (approved):

PR #435, opened 2023-08-24
use beacon api to send event depending on event type

olamothe:
This is a tentative fix for click events on iOS.

Discuss from where the issue originated: https://discuss.coveo.com/t/ua-drop-in-clickthrough-events-following-a-migration-from-jsui-to-atomic-headless/9958

Sending events when the end users is getting redirected (for example when they click on a result from a search page) will often fail on iOS + Safari. 

The reason is relatively simple to understand: When the current page is getting unloaded from memory from the browser, the browser won't necessarily wait to finish all current function execution, in order to speed up navigation for the end user.

The proposed fix is to use the navigator beacon API for click events specifically: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon

The beacon API  support is already implemented in the project, and seems to be working correctly.

Even with this change, if you look at browser dev tool on Safari + iOS, you'll see the beacon request often appear to fail (ie: cancelled).

However, with the manual tests I've done, when you actually look at the visit browser in the admin UI, the clicks appears correctly tracked on iOS Safari. 

It would be interesting to add some full end to end test for this, since we can't really rely on what's happening client side to verify that clicks were actually captured by the platform...

### DIFF
--- a/src/client/analyticsBeaconClient.ts
+++ b/src/client/analyticsBeaconClient.ts
@@ -26,7 +26,6 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
         };
 
         // tslint:disable-next-line: no-console
-        console.log(`Sending beacon for "${eventType}" with: `, JSON.stringify(payload));
         navigator.sendBeacon(url, body as any); // https://github.com/microsoft/TypeScript/issues/38715
         return;
     }
--- a/src/client/analytics.ts
+++ b/src/client/analytics.ts
@@ -393,7 +393,7 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
                     this.afterSendHooks.map((hook) => hook(eventType, {...parametersToSend, ...remainingPayload}))
                 );
                 await this.deferExecution();
-                return (await this.sendFromBufferWithFetch()) as TResponse | void;
+                return (await this.sendFromBuffer()) as TResponse | void;
             },
         };
     }
@@ -406,11 +406,11 @@ export class CoveoAnalyticsClient implements AnalyticsClient, VisitorIdProvider 
         return new Promise((resolve) => setTimeout(resolve, 0));
     }
 
-    private async sendFromBufferWithFetch(): Promise<AnyEventResponse | void> {
+    private async sendFromBuffer(): Promise<AnyEventResponse | void> {
         const popped = this.bufferedRequests.shift();
         if (popped) {
             const {eventType, payload} = popped;
-            return this.runtime.client.sendEvent(eventType, payload);
+            return this.runtime.getClientDependingOnEventType(eventType).sendEvent(eventType, payload);
         }
     }
 
--- a/src/client/runtimeEnvironment.ts
+++ b/src/client/runtimeEnvironment.ts
@@ -4,10 +4,12 @@ import {hasLocalStorage, hasCookieStorage} from '../detector';
 import {AnalyticsRequestClient, IAnalyticsClientOptions, NoopAnalyticsClient} from './analyticsRequestClient';
 import {AnalyticsFetchClient} from './analyticsFetchClient';
 import {BufferedRequest} from './analytics';
+import {EventType} from '../events';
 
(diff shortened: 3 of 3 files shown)


rphilippen:
> Even with this change, if you look at browser dev tool on Safari + iOS, you'll see the beacon request often appear to fail (ie: cancelled). So, all in all, Safari is buggy/unreliable regarding all of this.

When Frank and I looked into a different thing related to the beacon API in Chrome, we observed a similar thing.
From MDN:
> The Beacon API is used to send an asynchronous and non-blocking request to a web server. 

That means these requests are fire-and-forget to the page that initiated them, and I think browsers don't tie these requests to the page context in any way (so they can proceed unloading the page and all its related artifacts, such as scripts). This is speculation on my part, but I think the dev tools simply get disconnected from the page before the request finishes, and thus show the request as failed because of that.

src/client/analyticsBeaconClient.ts
```diff
@@ -26,11 +26,14 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
         };
 
         // tslint:disable-next-line: no-console
```
rphilippen, line 28: Since you removed the following console.log statement, you should also remove this tslint pragma to not warn about it.
```suggestion
```

src/client/analyticsBeaconClient.ts
```diff
@@ -26,11 +26,14 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
...
-        console.log(`Sending beacon for "${eventType}" with: `, JSON.stringify(payload));
         navigator.sendBeacon(url, body as any); // https://github.com/microsoft/TypeScript/issues/38715
         return;
     }
 
+    public isAvailable() {
+        return typeof navigator.sendBeacon === 'function';
```
rphilippen, line 34: It's funny how I still have a trigger for this, because of Internet Explorer (shows my age) that would report `object` instead of `function` for built-in methods on the web API.

That aside, I do wonder why you use this stricter check over the `in` operator? It's what MDN suggest for [Implementing feature detection](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection#javascript).
```suggestion
        return 'sendBeacon' in navigator;
```

rphilippen (approved):

fvanham-coveo:
This PR is actually quite impactful, in the sense that it'll move ALL clicks (only ua clicks, but not collect clicks which are also used) to beacon, which means we lose visibility on their payloads. This will make it really hard to debug click events in any Coveo integration. Secondly, we should have a mechanism to send all "lagging" events, irrespective of their types.

Instead I'm wondering if the trigger we are using for sending "backlogged" analytics events is simply too unreliable: we're using `beforeunload` in [this hook](https://github.com/coveo/coveo.analytics.js/blob/0742718d3b66634b22ea6d0edfbd73be3bf4441c/src/client/runtimeEnvironment.ts#L29). Current docs are flagging this event as unreliable on (especially) mobile:
- https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event
- https://developer.chrome.com/articles/page-lifecycle-api/#developer-recommendations-for-each-state, which explicitly states "Never add a beforeunload listener unconditionally or use it as an end-of-session signal." (which is what we are doing).

I'll take a look at how reliable `pagehide` is on mobile.

fvanham-coveo (approved):
I'm going to approve these changes in order to fix an immediate client issue. Merging this will have the following consequences:
- This may only "fix" things for clicks. Any other straggling events will still be left behind once a page reloads. However, clicks are the main cause of page reloads so that's acceptable for a fix.
- We implicitly impact browser compatibility. IE and opera mini do not support Beacon. Most modern browsers do, so that's acceptable.

A lot of the logic flow inside coveo.analytics.js is async, and I suspect it gets cut off by the browser when the page is reloaded. We should see whether this fix improves things, and then rethink.

olamothe:
Update: @mrrajamanickam-coveo will take a crack at fixing the core issue this sprint.

PR #436, opened 2023-09-12
Fix preprocess request for send beacon

olamothe:
Following the changes to use `sendBeacon` in Headless for clicks (through coveo.analytics  PR: https://github.com/coveo/coveo.analytics.js/pull/435) , we've stumbled upon a problem in the `preprocess` functionality that is implemented in the Beacon client

It's a functionality that allows implementers to either augment some analytics event (add custom data), override properties, or simply use them as hooks to copy the events payload to other system (google analytics). They are often used in the field.

The main problem is that the old `preprocess` functionality for sendBeacon was sending the payload as a [Blob](https://developer.mozilla.org/en-US/docs/Web/API/Blob) to the implementer's `preprocess` function.

Even though the function signature says that the request body "could be anything" (including a Blob), the reality is that many implementers won't defensively code around that fact, and generally expect everything in there to be a JSON string.

Also, Blob in general are unwieldy to manipulate.

This PR changes this so that we send the stringified JSON object to the implementers (similar to how the `analyticsFetchClient` does it), before changing it to a blob and sending the event.

### DIFF
--- a/src/client/analyticsBeaconClient.spec.ts
+++ b/src/client/analyticsBeaconClient.spec.ts
@@ -162,6 +162,20 @@ describe('AnalyticsBeaconClient', () => {
             );
         });
 
+        it('to augment the request body as a JSON string for a click event', async () => {
+            const client = setupClient((request) => {
+                const bodyParsedAsJSON = JSON.parse(request.body as string);
+                bodyParsedAsJSON.aNewProperty = 'bar';
+                request.body = JSON.stringify(bodyParsedAsJSON);
+                return request;
+            });
+
+            await client.sendEvent(EventType.click, {actionCause: 'foo'});
+            expect(await getSendBeaconFirstCallBlobArgument()).toContain(
+                `clickEvent=${encodeURIComponent(`{"actionCause":"foo","aNewProperty":"bar"}`)}`
+            );
+        });
+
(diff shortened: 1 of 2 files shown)


src/client/analyticsBeaconClient.spec.ts
```diff
@@ -99,33 +99,80 @@ describe('AnalyticsBeaconClient', () => {
...
+            });
+
+            await client.sendEvent(EventType.click, {actionCause: 'foo'});
+            expect(await getSendBeaconFirstCallBlobArgument()).toContain(
+                `clickEvent=${encodeURIComponent(`{"actionCause":"bar"}`)}`
+            );
+        });
 
-        expect(clientOrigin!).toBe('analyticsBeacon');
-        expect(sendBeaconMock).toHaveBeenCalledWith(processedRequest.url, processedRequest.body);
```
olamothe, line 128: This is one test that needed to be modified (the expectation that the mock is called with `processedRequest.body`)

And to my understanding, it seems the old test was faulty: It expected the mock to have been called with the modified "raw" JSON object, as opposed to something that can actually be accepted by the beacon API (ie: a blob).

louis-bompart (approved):

aperron-coveo (approved):

PR #438, opened 2023-10-05
feat(genQA): new RGA UA events added to the search page client

mmitiche:
[SFINT-5206](https://coveord.atlassian.net/browse/SFINT-5206)

#### Custom events:

- generatedAnswerCitationHover
- generatedAnswerCopyToClipboard
- generatedAnswerHideAnswers
- generatedAnswerShowAnswers
- generativeQuestionFeedbackSubmit

#### Search events:

- rephraseGeneratedAnswer

#### example of adding the fields `showGeneratedAnswer` and `generatedAnswerFormat`: 

ExampleSearch

### The events are documented here: [UA events schemas- GenAI v2](https://docs.google.com/document/d/1vv9hKHZJ4tjSFRNoxG88DKqeTxvBAG-jBYMp-mAg6oE/edit#heading=h.91bl6jfm09i6)

[SFINT-5206]: https://coveord.atlassian.net/browse/SFINT-5206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/searchPage/searchPageClient.ts
+++ b/src/searchPage/searchPageClient.ts
@@ -39,6 +39,9 @@ import {
     GeneratedAnswerFeedbackMeta,
     GeneratedAnswerCitationMeta,
     GeneratedAnswerStreamEndMeta,
+    GeneratedAnswerCitationHoverMeta,
+    GeneratedAnswerBaseMeta,
+    GeneratedAnswerRephraseMeta,
 } from './searchPageEvents';
 import {NoopAnalytics} from '../client/noopAnalytics';
 import {formatOmniboxMetadata} from '../formatting/format-omnibox-metadata';
@@ -885,19 +888,19 @@ export class CoveoSearchPageClient {
         };
     }
 
-    public makeLikeGeneratedAnswer(metadata: GeneratedAnswerFeedbackMeta) {
+    public makeLikeGeneratedAnswer(metadata: GeneratedAnswerBaseMeta) {
         return this.makeCustomEvent(SearchPageEvents.likeGeneratedAnswer, metadata);
     }
 
-    public async logLikeGeneratedAnswer(metadata: GeneratedAnswerFeedbackMeta) {
+    public async logLikeGeneratedAnswer(metadata: GeneratedAnswerBaseMeta) {
         return (await this.makeLikeGeneratedAnswer(metadata)).log({searchUID: this.provider.getSearchUID()});
     }
 
-    public makeDislikeGeneratedAnswer(metadata: GeneratedAnswerFeedbackMeta) {
+    public makeDislikeGeneratedAnswer(metadata: GeneratedAnswerBaseMeta) {
         return this.makeCustomEvent(SearchPageEvents.dislikeGeneratedAnswer, metadata);
     }
 
-    public async logDislikeGeneratedAnswer(metadata: GeneratedAnswerFeedbackMeta) {
+    public async logDislikeGeneratedAnswer(metadata: GeneratedAnswerBaseMeta) {
         return (await this.makeDislikeGeneratedAnswer(metadata)).log({searchUID: this.provider.getSearchUID()});
     }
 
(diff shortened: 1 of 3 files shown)


SimonMilord (approved):
👌

src/searchPage/searchPageEvents.ts
```diff
@@ -509,3 +536,20 @@ export interface GeneratedAnswerCitationMeta {
     permanentId: string;
     id: string;
 }
+
+export type GeneratedAnswerFeedbackReason = 'irrelevant' | 'notAccurate' | 'outOfDate' | 'harmful' | 'other';
+
+export type GeneratedAnswerRephraseStrategy = 'stepByStep' | 'bulletPoints' | 'summarize';
+
+export interface GeneratedAnswerCitationHoverMeta extends GeneratedAnswerCitationMeta {
```
erocheleau, line 544: Wouldn't we need something to identity which citation was hovered on also?

Either the permanentId or if the citation has a citationId?

src/searchPage/searchPageClient.ts
```diff
@@ -911,6 +914,64 @@ export class CoveoSearchPageClient {
         });
     }
 
+    public makeGeneratedAnswerCitationHover(metadata: GeneratedAnswerCitationHoverMeta) {
```
erocheleau, line 917: makeGeneratedAnswerSourceHover?

src/searchPage/searchPageClient.ts
```diff
@@ -911,6 +914,64 @@ export class CoveoSearchPageClient {
...
+        return this.makeCustomEvent(SearchPageEvents.generatedAnswerCitationHover, metadata);
+    }
+
+    public async logGeneratedAnswerCitationHover(metadata: GeneratedAnswerCitationHoverMeta) {
```
erocheleau, line 921: logGeneratedAnswerSourceHover?

src/searchPage/searchPageEvents.ts
erocheleau: Nvm me it's in the extended interface ;)

src/searchPage/searchPageEvents.ts
```diff
@@ -314,6 +314,30 @@ export enum SearchPageEvents {
      * Identified the custom event that gets logged when a generated answer stream is completed.
      */
     generatedAnswerStreamEnd = 'generatedAnswerStreamEnd',
+    /**
+     * Identifies the custom event that gets logged when a user hovers over a generated answer citation.
+     */
+    generatedAnswerCitationHover = 'generatedAnswerCitationHover',
```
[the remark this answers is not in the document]
mmitiche: Yep will update that 👌

src/searchPage/searchPageEvents.ts
```diff
@@ -509,3 +536,20 @@ export interface GeneratedAnswerCitationMeta {
     permanentId: string;
     id: string;
 }
+
+export type GeneratedAnswerFeedbackReason = 'irrelevant' | 'notAccurate' | 'outOfDate' | 'harmful' | 'other';
+
+export type GeneratedAnswerRephraseStrategy = 'stepByStep' | 'bulletPoints' | 'summarize';
```
nathanlb, line 542:
```suggestion
export type GeneratedAnswerRephraseFormat = 'stepByStep' | 'bulletPoints' | 'summarize';
```
I believe "format" is more used than "strategy"

src/searchPage/searchPageEvents.ts
```diff
@@ -509,3 +536,20 @@ export interface GeneratedAnswerCitationMeta {
...
+
+export interface GeneratedAnswerCitationHoverMeta extends GeneratedAnswerCitationMeta {
+    citationHoverTime: number;
+}
+
+export interface GeneratedAnswerRephraseMeta extends GeneratedAnswerBaseMeta {
+    rephraseStrategy: GeneratedAnswerRephraseStrategy;
```
nathanlb, line 549:
```suggestion
    generatedAnswerFormat: GeneratedAnswerRephraseFormat;
```
Seems to be format in the google doc

src/searchPage/searchPageEvents.ts
```diff
@@ -509,3 +536,20 @@ export interface GeneratedAnswerCitationMeta {
     permanentId: string;
     id: string;
```
felixpelletier, line 537: Maybe it's too late, but in smart snippet we seem to call that documentId (or is that now the same thing?)

src/searchPage/searchPageEvents.ts
jstpierref: `GeneratedAnswerFormat` to match the updated design doc

fvanham-coveo (approved):

jmprovencher:
Don't we think that generatedAnswerCitationHover might generate a little bit too much data and does it actually provides us with any clear insight ?

nathanlb:
> Don't we think that generatedAnswerCitationHover might generate a little bit too much data and does it actually provides us with any clear insight ?

It's there because hovering over the citation will show a popup with the relevant document snippet inside and we want to see if it is read.
We'll put it on a debounce timer so it should only trigger an event when the hovering lasted a second or two.

jmprovencher:
> > Don't we think that generatedAnswerCitationHover might generate a little bit too much data and does it actually provides us with any clear insight ?
> 
> It's there because hovering over the citation will show a popup with the relevant document snippet inside and we want to see if it is read. We'll put it on a debounce timer so it should only trigger an event when the hovering lasted a second or two.

Should we allow more than one hover events per generated answer?

jmprovencher (approved):

src/searchPage/searchPageEvents.ts
```diff
@@ -509,3 +536,20 @@ export interface GeneratedAnswerCitationMeta {
...
 }
+
+export type GeneratedAnswerFeedbackReason = 'irrelevant' | 'notAccurate' | 'outOfDate' | 'harmful' | 'other';
+
+export type GeneratedAnswerRephraseFormat = 'step' | 'bullet' | 'concise' | 'default';
```
jstpierref, line 542: This will already be part of the SAPI request (therefore logged as SSE), but I'm not against adding it client side.

nathanlb:
> > > Don't we think that generatedAnswerCitationHover might generate a little bit too much data and does it actually provides us with any clear insight ?
> > 
> > 
> > It's there because hovering over the citation will show a popup with the relevant document snippet inside and we want to see if it is read. We'll put it on a debounce timer so it should only trigger an event when the hovering lasted a second or two.
> 
> Should we allow more than one hover events per generated answer?

Each generated answer can have up to 5 citations and each citation will be hover-able. I don't think capping it makes sense. Given the hover needs to be for a "meaningful" amount of time for the event to be sent, I doubt it'll be accidentally abused.

jstpierref (approved):

src/searchPage/searchPageEvents.ts
erocheleau: We have a custom event when someone clicks on the reformat buttons, so it's for a different event than the search one.

erocheleau (approved):

src/searchPage/searchPageEvents.ts
felixpelletier: My bad: permanentId is fine here, but I think id should be called `citationId`. It's not always clear what id this is.

mmitiche: The thing is that we are already logging the `OpenGeneratedAnswerSource` event with the `id` and not `citationId` so I think it's too late to change this now, it's kinda a small breaking change.

felixpelletier:
> > > > Don't we think that generatedAnswerCitationHover might generate a little bit too much data and does it actually provides us with any clear insight ?
> > > 
> > > 
> > > It's there because hovering over the citation will show a popup with the relevant document snippet inside and we want to see if it is read. We'll put it on a debounce timer so it should only trigger an event when the hovering lasted a second or two.
> > 
> > 
> > Should we allow more than one hover events per generated answer?
> 
> Each generated answer can have up to 5 citations and each citation will be hover-able. I don't think capping it makes sense. Given the hover needs to be for a "meaningful" amount of time for the event to be sent, I doubt it'll be accidentally abused.

Yep. To be clear, we're looking at new ways to track whether customers "click" on documents (because they might not not that there's answers). We think that customers hovering of citations may be a sign of the customer needing more insight, but the info in the citation hover being good enough (there's nothing in the citation hover today, but there could very well be. We heard that as an idea several times). Could be a cue to indicate the answer is suboptimal but the content is good.

src/searchPage/searchPageEvents.ts
felixpelletier: Yeah, but we aren't using it yet as far as I know, so it's probably the best time to change it.

felixpelletier: I just asked Ezmie. She too think it's best to do it now.

felixpelletier (approved):

nathanlb (approved):

PR #439, opened 2023-10-06
Re-added old bundles

btaillon-coveo:
# Context

## Before the breaking change
Prior to https://github.com/coveo/coveo.analytics.js/pull/433, it was not fully possible to import directly from `coveo.analytics` like so:
```js
import {CoveoAnalyticsClient} from 'coveo.analytics';
```

This was because the `package.json` file contained these lines:
```json
{
    "main": "dist/library.js",
    "module": "dist/library.es.js",
    "browser": "dist/coveoua.browser.js",
}
```

While `library.js` and `library.es.js` contained modules, `coveoua.browser.js` did not (it globally defines "coveoua"), and the "browser" entry caused some bundlers to try and fail to use it. Additionally, `library.es.js` was using expecting browser APIs while `library.js` was expecting Node.js APIs.

## The breaking change
https://github.com/coveo/coveo.analytics.js/pull/433 introduced a way to import from `coveo.analytics` like so:
```js
import {CoveoAnalyticsClient} from 'coveo.analytics/modules';
```

In that process, to avoid confusion between Node CommonJS, Node ESM, browser ESM, and browser IIFE, I:
- renamed `library.js` to `library.cjs`
- added `library.esm`
- renamed `library.es.js` to `browser.mjs`

Unfortunately, some of our own implementations (and likely some external) were using a workaround which involved forcefully telling their bundlers to use `library.es.js`. Since this file was renamed, those now crash.

# Solution

I added copies of the renamed files to their former names.

Specifically:
- `browser.mjs` gets copied to `library.es.js`
- `library.cjs` gets copied to `library.js`


alexprudhomme (approved):
Nice solution !

btaillon-coveo:
confirmed that this works by testing with `npm link` and a separate repo.

fbeaudoincoveo (approved):

PR #440, opened 2023-10-10
feat(genQa): RGA ua events for insight use case created

mmitiche:
[SFINT-5198](https://coveord.atlassian.net/browse/SFINT-5198)

#### Custom events:

- likeGeneratedAnswer
- dislikeGeneratedAnswer
- openGeneratedAnswerSource
- generatedAnswerStreamEnd
- generatedAnswerSourceHover
- generatedAnswerCopyToClipboard
- generatedAnswerHideAnswers
- generatedAnswerShowAnswers
- generativeQuestionFeedbackSubmit

#### Search events:

- retryGeneratedAnswer
- rephraseGeneratedAnswer

### Added the ability to send generated answer metadata in all the Search UA events by adding a new function `getGeneratedAnswerMetadata` to the `InsightClientProvider`

[SFINT-5198]: https://coveord.atlassian.net/browse/SFINT-5198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/insight/insightClient.spec.ts
+++ b/src/insight/insightClient.spec.ts
@@ -600,7 +600,7 @@ describe('InsightClient', () => {
         it('should send proper payload for #rephraseGeneratedAnswer', async () => {
             const exampleGeneratedAnswerMetadata = {
                 generativeQuestionAnsweringId: '123',
-                rephraseFormat: <GeneratedAnswerRephraseFormat>'stepByStep',
+                rephraseFormat: <GeneratedAnswerRephraseFormat>'step',
             };
 
             await client.logRephraseGeneratedAnswer(exampleGeneratedAnswerMetadata);
@@ -1346,7 +1346,7 @@ describe('InsightClient', () => {
         it('should send proper payload for #rephraseGeneratedAnswer', async () => {
             const exampleGeneratedAnswerMetadata = {
                 generativeQuestionAnsweringId: '123',
-                rephraseFormat: <GeneratedAnswerRephraseFormat>'stepByStep',
+                rephraseFormat: <GeneratedAnswerRephraseFormat>'step',
             };
             const expectedMetadata = {
                 ...exampleGeneratedAnswerMetadata,
--- a/src/insight/insightClient.spec.ts
+++ b/src/insight/insightClient.spec.ts
@@ -65,6 +65,7 @@ describe('InsightClient', () => {
         getLanguage: () => 'en',
         getFacetState: () => fakeFacetState,
         getIsAnonymous: () => false,
+        getGeneratedAnswerMetadata: () => ({genratedAnswerMetadata: 'foo'}),
     };
 
     const initClient = () => {
@@ -80,7 +81,7 @@ describe('InsightClient', () => {
 
     const expectMatchPayload = (actionCause: SearchPageEvents | InsightEvents, meta = {}) => {
         const body: string = lastCallBody(fetchMock);
-        const customData = {foo: 'bar', ...meta};
+        const customData = {foo: 'bar', genratedAnswerMetadata: 'foo', ...meta};
         expect(JSON.parse(body)).toMatchObject({
             queryText: 'queryText',
             responseTime: 123,
--- a/src/insight/insightClient.spec.ts
+++ b/src/insight/insightClient.spec.ts
@@ -535,7 +535,7 @@ describe('InsightClient', () => {
             const exampleGeneratedAnswerMetadata = {
                 generativeQuestionAnsweringId: '123',
                 permanentId: 'foo',
-                id: 'bar',
+                citationId: 'bar',
             };
 
             await client.logOpenGeneratedAnswerSource(exampleGeneratedAnswerMetadata);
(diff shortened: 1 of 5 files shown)


src/insight/insightClient.ts
```diff
@@ -476,6 +483,121 @@ export class CoveoInsightClient {
...
+    public logGeneratedAnswerShowAnswers(generatedAnswerMetadata: GeneratedAnswerBaseMeta, metadata?: CaseMetadata) {
+        return this.logCustomEvent(
+            SearchPageEvents.generatedAnswerShowAnswers,
+            metadata
+                ? {...generateMetadataToSend(metadata, false), ...generatedAnswerMetadata}
+                : generatedAnswerMetadata
+        );
+    }
+
+    public logGenerativeQuestionFeedbackSubmit(
```
erocheleau, line 558:
```suggestion
    public logGeneratedAnswerFeedbackSubmit(
```
???

src/insight/insightClient.ts
```diff
@@ -476,6 +483,121 @@ export class CoveoInsightClient {
...
+        generatedAnswerFeedbackMetadata: GeneratedAnswerFeedbackMeta,
+        metadata?: CaseMetadata
+    ) {
+        return this.logCustomEvent(
+            SearchPageEvents.generativeQuestionFeedbackSubmit,
```
erocheleau, line 563: ugh we have only this one who is not named like everything else? :(

erocheleau: If that's the case I prefer at least the `logX` methods use all the right name and hide this discrepancy .

src/insight/insightClient.ts
```diff
@@ -476,6 +483,121 @@ export class CoveoInsightClient {
...
+            metadata
+                ? {...generateMetadataToSend(metadata, false), ...generatedAnswerFeedbackMetadata}
+                : generatedAnswerFeedbackMetadata
+        );
+    }
+
+    public logRephraseGeneratedAnswer(
```
erocheleau, line 570:
```suggestion
    public logGeneratedAnswerRephrase(
```
To fit with all the others?

src/insight/insightClient.ts
```diff
@@ -476,6 +483,121 @@ export class CoveoInsightClient {
...
+    ) {
+        return this.logSearchEvent(
+            SearchPageEvents.rephraseGeneratedAnswer,
+            metadata
+                ? {...generateMetadataToSend(metadata, false), ...generatedAnswerRephraseMetadata}
+                : generatedAnswerRephraseMetadata
+        );
+    }
+
+    public logRetryGeneratedAnswer(metadata?: CaseMetadata) {
```
erocheleau, line 582:
```suggestion
    public logGeneratedAnswerRetry(metadata?: CaseMetadata) {
```
?

erocheleau (commented):
Some suggest names change, did we mess up those names in the previous PR ? :(

src/insight/insightClient.spec.ts
```diff
@@ -509,6 +512,116 @@ describe('InsightClient', () => {
...
+        it('should send proper payload for #generatedAnswerShowAnswers', async () => {
+            const exampleGeneratedAnswerMetadata = {
+                generativeQuestionAnsweringId: '123',
+            };
+
+            await client.logGeneratedAnswerShowAnswers(exampleGeneratedAnswerMetadata);
+            expectMatchCustomEventPayload(SearchPageEvents.generatedAnswerShowAnswers, exampleGeneratedAnswerMetadata);
+        });
+
+        it('should send proper payload for #generativeQuestionFeedbackSubmit', async () => {
```
SimonMilord, line 587:
```suggestion
        it('should send proper payload for #generatedAnswerFeedbackSubmit', async () => {
```

should probably change these if we go with @erocheleau suggestions

src/insight/insightClient.spec.ts
```diff
@@ -509,6 +512,116 @@ describe('InsightClient', () => {
...
+            const exampleGeneratedAnswerMetadata = {
+                generativeQuestionAnsweringId: '123',
+                reason: <GeneratedAnswerFeedbackReason>'other',
+                details: 'foo',
+            };
+
+            await client.logGenerativeQuestionFeedbackSubmit(exampleGeneratedAnswerMetadata);
```
SimonMilord, line 594: same here

SimonMilord (approved):
I agree with etienne's suggestions and if we decide to go with that naming then tests naming should also be updated. Other than that it looks fine to me

mmitiche:
> Some suggest names change, did we mess up those names in the previous PR ? :(

yep these were the name we used in the previous PR and the names that were mentioned in the design doc 🤷‍♂️ do we still wanna change ? x)

src/insight/insightClient.ts
mmitiche: This event has been updated for both the search and insight use cases

src/insight/insightClient.ts
mmitiche: I don't think we have a rule about that, looking at all the events we currently have we are using these two patterns interchangeably, thus I don't think this change is worth the trouble especially that  both `logRephraseGeneratedAnswer` and `logGeneratedAnswerRephrase` make sense.

erocheleau (approved):

PR #445, opened 2023-11-21
feat(create article): UA event added for createArticle

SimonMilord:
[SFINT-5276](https://coveord.atlassian.net/browse/SFINT-5276)

**The following custom event has been added to Insight Client for knowledge article creation:**

- createArticle

[SFINT-5276]: https://coveord.atlassian.net/browse/SFINT-5276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/searchPage/searchPageEvents.ts
```diff
@@ -389,6 +389,7 @@ export const CustomEventsTypes: Partial<Record<SearchPageEvents | InsightEvents,
     [SearchPageEvents.generatedAnswerHideAnswers]: 'generatedAnswer',
     [SearchPageEvents.generatedAnswerShowAnswers]: 'generatedAnswer',
     [SearchPageEvents.generatedAnswerFeedbackSubmit]: 'generatedAnswer',
+    [InsightEvents.createArticle]: 'createArticle',
```
erocheleau, line 392: Is there any kind of order here? I would group it with the other InsightEvents :D

erocheleau (approved):
looks good

src/searchPage/searchPageEvents.ts
SimonMilord: I thought newer events at the bottom, but I can move it 😂

mmitiche (changes requested):
In the Create Article event currently sent in the standard agent panel, the event contains the following custom data:

Screenshot 2023-11-22 at 7 08 12 AM

Notice the fields `articleType` and `triggeredBy` are being sent in addition to the context.

Please consider adding these two fields to the metadata accepted by the `logCreateArticle` function added in this PR.

erocheleau:
> In the Create Article event currently sent in the standard agent panel, the event contains the following custom data:
> 
> Screenshot 2023-11-22 at 7 08 12 AM
> Notice the fields `articleType` and `triggeredBy` are being sent in addition to the context.
> 
> Please consider adding these two fields to the metadata accepted by the `logCreateArticle` function added in this PR.

I would have to go verify, but I don't think we know the article type, do we? xD 
We used to have an option to specify the article type but we won't be able to have that in the HIP at least not yet.

src/insight/insightClient.ts
```diff
@@ -266,6 +267,13 @@ export class CoveoInsightClient {
         });
     }
 
+    public logCreateArticle(createArticle: CreateArticleMetadata, metadata?: CaseMetadata) {
```
mmitiche, line 270:
```suggestion
    public logCreateArticle(createArticleMetadata: CreateArticleMetadata, metadata?: CaseMetadata) {
```

src/insight/insightClient.ts
```diff
@@ -266,6 +267,13 @@ export class CoveoInsightClient {
...
+        return this.logCustomEvent(
+            InsightEvents.createArticle,
+            metadata ? {...generateMetadataToSend(metadata, false), ...createArticle} : {createArticle}
```
mmitiche, line 273: `createArticleMetadata` is already an object, we would get a different structure in the event custom data if we keep it `{createArticle}`
```suggestion
            metadata ? {...generateMetadataToSend(metadata, false), ...createArticle} : createArticle
```

src/insight/insightClient.ts
SimonMilord: I was using the same pattern as :

`logPageViewClick(pageView: UserActionsPageViewMetadata, metadata: CaseMetadata)`

but sure I can change it

mmitiche (approved):
LGTM 👌

PR #449, opened 2024-01-04
[UA-8522] Encode payload as json rather than form

fvanham-coveo:
This change modifies the way the beacon client sends information to the UA endpoint. UA endpoint has already been modified in a backwards compatible way in [this change set](https://github.com/coveo/usageanalytics/pull/2686). Prior to the below change coveoua was sending in collect events in a form-encoded way, where each parameter is a separate URI encoded entry.
```
access_token=TOK&pa=add&pr1id=SKU%202353&tm=1243345
```
When the write service receives this content, it has no other option than to treat all parameters as strings and logs them in the data lake as such, which causes issues with validating parameters that were originally supplied as numeric (e.g. 1243345 above).

After this change, coveoua will still send in the information as formencoded, but the entire event content is stringified JSON. This allows re-parsing of the json on the backend, which can use double quotes to identify strings vs numeric variables:
```
access_token=TOK&collectEvent=%7B%22pa%22%3A%22add%22...%7D
```
The access_token is still part of the form parameters, as the original endpoint for receiving beaconCollect events requires it to be, and removing it would break existing logging with older versions of coveoua. Technically it could also have been sent as a query param. Note that we cannot introduce a collect parameter called `collectEvent` anymore, since the server side logic switches based on the presence of that formParameter.

I modified the index.html file to allow beacon testing. I'm not sure how this was tested previously. I verified that events logged using the beacon api with the new format end up in the data lake as expected.

---
Note that this change is hard to verify, since there's no good way to verify that beacon-collect events are logged appropriately in production (since they are quite rare), and it'll be impossible to detect that we are dropping these type of events on the backend. Prior to deployment the following steps should be taken:
- Verify that the write service change has been deployed in prod (!)
- Use the index.html page to manually send a few beacon-collect events to UA prod and verify they are logged in RAW
- Use a redirector plugin to load the new version of coveoua into a customer integration and verify existing logging still works: beacon-click, standard collect. It'll be very hard to trigger a beacon-collect event with site interactions.

### DIFF
--- a/src/client/analyticsBeaconClient.ts
+++ b/src/client/analyticsBeaconClient.ts
@@ -68,8 +68,8 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
 
     private encodeForEventType(eventType: EventType, payload: IRequestPayload): string {
         return this.isEventTypeLegacy(eventType)
-            ? this.encodeToJson(eventType, payload)
-            : this.encodeToJson(eventType, payload, this.opts.token);
+            ? this.encodeEventToJson(eventType, payload)
+            : this.encodeEventToJson(eventType, payload, this.opts.token);
     }
 
     private async getQueryParamsForEventType(eventType: EventType): Promise<string> {
@@ -88,13 +88,11 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
         return [EventType.click, EventType.custom, EventType.search, EventType.view].indexOf(eventType) !== -1;
     }
 
-    private encodeToJson(eventType: EventType, payload: IRequestPayload, access_token?: string): string {
+    private encodeEventToJson(eventType: EventType, payload: IRequestPayload, access_token?: string): string {
+        let encoded = `${eventType}Event=${encodeURIComponent(JSON.stringify(payload))}`;
         if (access_token) {
-            return `access_token=${encodeURIComponent(access_token)}&${eventType}Event=${encodeURIComponent(
-                JSON.stringify(payload)
-            )}`;
-        } else {
-            return `${eventType}Event=${encodeURIComponent(JSON.stringify(payload))}`;
+            encoded = `access_token=${encodeURIComponent(access_token)}&${encoded}`;
         }
+        return encoded;
     }
 }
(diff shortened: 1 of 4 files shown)


aperron-coveo (approved):

src/client/analyticsBeaconClient.ts
```diff
@@ -91,20 +88,13 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
...
-
-    private encodeValue(value: any) {
-        return typeof value === 'number' || typeof value === 'string' || typeof value === 'boolean'
-            ? value
-            : JSON.stringify(value);
+    private encodeToJson(eventType: EventType, payload: IRequestPayload, access_token?: string): string {
+        if (access_token) {
+            return `access_token=${encodeURIComponent(access_token)}&${eventType}Event=${encodeURIComponent(
+                JSON.stringify(payload)
+            )}`;
```
rphilippen, line 95: Took me a second to realize these newlines and indenting whitespace don't end up in the final string, because it's part of the `encodeURIComponent` arguments 😅.

As a more general remark, there is quite a large repeated part. It might make the code shorter and easier to comprehend to take that out instead? I'm also not 100% on board with the function name "encodeToJson", as this method actually encodes the event to query parameters. I'm not going to change it for my "suggestion" though, as it would make it less clear what the "diff" is.

```typescript
    private encodeToJson(eventType: EventType, payload: IRequestPayload, access_token?: string): string {
        let encoded = `${eventType}Event=${encodeURIComponent(JSON.stringify(payload))}`;
        if (access_token) {
            encoded = `access_token=${encodeURIComponent(access_token)}&${encoded}`;
        }
        return encoded;
    }
```

rphilippen (approved):

olamothe (approved):

src/client/analyticsBeaconClient.ts
```diff
@@ -91,20 +88,13 @@ export class AnalyticsBeaconClient implements AnalyticsRequestClient {
...
-            .join('&');
-    }
-
-    private encodeValue(value: any) {
-        return typeof value === 'number' || typeof value === 'string' || typeof value === 'boolean'
-            ? value
-            : JSON.stringify(value);
+    private encodeToJson(eventType: EventType, payload: IRequestPayload, access_token?: string): string {
+        if (access_token) {
+            return `access_token=${encodeURIComponent(access_token)}&${eventType}Event=${encodeURIComponent(
```
olamothe, line 93: Can access_token really be null/undefined here ?

If so, there should be some defensive code, if just to ensure we don't send `access_token=undefined`

fvanham-coveo: The irony is access token can be undefined for the case of legacy events. In that case the access token is in the query param and not in the form field.

PR #454, opened 2024-03-20
[UA-8748] Adjust DNT logic

fvanham-coveo:
The existing logic for Do Not Track (DNT) in the basic coveoua AnalyticsClient was incorrect. The desired logic is:
- When a DNT header is present, no analytics events should be sent to the Coveo backend.
- When a DNT header is present, no cookies should be stored.
In collaboration with legal, we removed the requirement to *remove* already existing Cookies when a DNT header is found. There is no legal basis for this, and hidden removal of Cookies by software may be seen as intrusive.

The existing logic only removed the cookie, but did not disable the sending of analytics. This causes any integration using `coveoua` directly AND having a DNT header to initialize with a fresh clientId, but still send events. In practice this means, every time coveoua is initialized on a fresh page, the event will have a new clientID for clients with a DNT header. Needless to say, this causes bad downstream effects.


jmprovencher (approved):

samisayegh (approved):
✅

PR #458, opened 2024-05-17
fix(ec): pageview will not include action, product and impression

sallainCoveo:
Jira: https://coveord.atlassian.net/browse/LENS-1826

The issue described in the Jira, Slack thread and discuss was that sometimes the "event" type event didn't include the product and product action / the "pageview" type event include the product and product action.

It would seems that the order of event that are sent is important, but to prevent the plugin's data to be cleared too soon OR that we include the data to a `pageview` event, I've added a pageview check.

One argument that could be made is to put the emphasis in the documentation that the "event" type should absolutely be sent before the pageview to prevent this problem. However, there's also an argument to be made that maybe the order shouldn't matter (especially when these events were historically coupled together, so more client's implementations might be broken then we think). 

Demo of the results:

https://github.com/coveo/coveo.analytics.js/assets/58052881/3d115b34-660a-421f-9e65-5d8183259ae5

### DIFF
--- a/functional/ec-events.spec.ts
+++ b/functional/ec-events.spec.ts
@@ -99,7 +99,7 @@ describe('ec events', () => {
         });
     });
 
-    it('can send a product detail view event', async () => {
+    it('will not include the product or action when sending a pageview event', async () => {
         coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', unknown: 'ok'});
         coveoua('ec:setAction', 'detail', {storeid: 'amazing'});
         await coveoua('send', 'pageview');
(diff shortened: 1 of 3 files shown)


aperron-coveo (approved):

src/plugins/ec.ts
```diff
@@ -135,21 +135,31 @@ export class ECPlugin extends BasePlugin {
     }
 
     private addECDataToPayload(eventType: string, payload: any) {
+        const isPageView = eventType === ECPluginEventTypes.pageview;
         const ecPayload = {
```
samisayegh, line 139: Nit: the current approach works, but it adds multiple `if` statements to change the behavior of the function, which makes it take longer to understand.

What if instead we used a single `if` statement, returning early in the case of a pageview? e.g.,

```
if (isPageview) {
  return {
     ...this.getLocationInformation(eventType, payload),
     ...this.getDefaultContextInformation(eventType),
     ...payload
  }
}

... same as before ...
```

samisayegh (approved):
Nice investigation and fix ✅

src/plugins/ec.ts
sallainCoveo: Good idea, will change!

functional/ec-events.spec.ts
```diff
@@ -109,6 +109,42 @@ describe('ec events', () => {
         expect(body).toEqual({
             ...defaultContextValues,
             t: 'pageview',
+        });
+    });
+
+    it('will include the product and action only when sending a detail event', async () => {
+        coveoua('ec:addProduct', {name: 'wow', id: 'something', brand: 'brand', unknown: 'ok'});
+        coveoua('ec:setAction', 'detail', {storeid: 'amazing'});
+        await coveoua('send', 'pageview');
```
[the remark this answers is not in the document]
sallainCoveo: I don't think `await` are required to reproduce the bug. In the demo in the description, there's no await, and regardless on how the `pageview` is placed, it will not have the product action/ product data attached to it.

The tricky part with these tests is to get the expected body (in a predictable order and moment), hence the `await`. However, I wonder if I could change the condition to simply check if the expected bodies are contained in the array instead :eyes:, I'll test this one out

PR #459, opened 2024-05-30
feat(genQA): add new version of UA events for feedback submit

jelmedini:
[SVCC-3816](https://coveord.atlassian.net/browse/SVCC-3816)

Add new analytic event for Feedback submit `logGeneratedAnswerFeedbackSubmitV2` to the search and insight clients.

This V2 event is added for the improvement of the current feedback modal.

[SVCC-3816]: https://coveord.atlassian.net/browse/SVCC-3816?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/insight/insightClient.spec.ts
+++ b/src/insight/insightClient.spec.ts
@@ -4,6 +4,7 @@ import {NoopAnalytics} from '../client/noopAnalytics';
 import {
     CustomEventsTypes,
     GeneratedAnswerFeedbackReason,
+    GeneratedAnswerFeedbackReasonOption,
     GeneratedAnswerRephraseFormat,
     PartialDocumentInformation,
     SearchPageEvents,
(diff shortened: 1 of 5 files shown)


src/insight/insightClient.spec.ts
```diff
@@ -1399,6 +1400,28 @@ describe('InsightClient', () => {
             expectMatchCustomEventPayload(SearchPageEvents.generatedAnswerFeedbackSubmit, expectedMetadata);
         });
 
+        it('should send proper payload for #generatedAnswerFeedbackSubmitV2', async () => {
```
mmitiche, line 1403: We usually add a test as well for the situation where this event is logged without the case meta data.
It should be added in the `describe('when the case metadata is not included',() =>{...})` present in Line 159 of this file.

mmitiche (approved):
Looks good to me ✅
Just suggested the addition of one unit test.

src/insight/insightClient.spec.ts
jelmedini: ooh right! thank you

src/searchPage/searchPageClient.spec.ts
```diff
@@ -1601,6 +1602,23 @@ describe('SearchPageClient', () => {
         expectMatchDescription(built.description, SearchPageEvents.generatedAnswerCollapse, meta);
     });
 
+    it('should send proper payload for #logGeneratedAnswerFeedbackSubmiV2', async () => {
```
LucaCogliandroCoveo, line 1605:
```suggestion
    it('should send proper payload for #logGeneratedAnswerFeedbackSubmitV2', async () => {
```
i think you're missing a "t" here

src/insight/insightClient.ts
```diff
@@ -590,6 +591,18 @@ export class CoveoInsightClient {
         );
     }
 
+    public logGeneratedAnswerFeedbackSubmitV2(
```
erocheleau, line 594: I really like that you made a V2 instead of making the V1 or V2 a parameter

erocheleau (approved):
lgtm

PR #471, opened 2024-07-31
fix: UA-9204 Safer URL truncation

rphilippen:
### Proposed changes:

Over a year ago, PR #405 added truncation to this library. The motivation: the URL gets truncated server-side anyway, so might as well lower the request size and not send data this will be removed anyway. A reviewer points out that it may lead to invalid URIs. The risk of this was deemed low.

Fair enough, it took quite some time, but customers are getting data health warnings about invalid URLs. The URLs in question are truncated at an unfortunate point, leaving part of an `%FF` escape sequence (e.g. `%9` or `%`) at the end of the URL. Long story short, this integration has URLs in an Asian language, and the URL standard mandates that any code point above 0xFF is encoded as utf-8 and embedded as escaped hex sequences. To illustrate, ツ encodes to `%E3%83%84` (this code point requires 3 utf-8 bytes).

The quick and easy fix would be to simply detect a possible incomplete % sequence at the end, but that would still have the potential of truncating within a utf-8 multibyte code point, leaving it broken. I decided to go the extra mile and detect that too, and truncate the whole multibyte sequence. That effectively means URLs may get truncated to up to 11(!) characters less than the limit, as the longest 4-byte utf-8 sequence takes 12 URL characters to encode.

#### Validating using native APIs (e.g. `URL`)

In the mentioned PR, somebody suggested using the [URL constructor](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) to validate if the URL is valid. They didn't offer a suggested recovery on failure, but the most sensible is truncating one character more and trying again.

| Detection | Incomplete escape sequence | Incomplete utf-8 |
|------------|----------------------|--------------------|
| `new URL()` | ❌ | ❌ |
| `URL.canParse()` | ❌ | ❌ |
| `decodeURIComponent()` | ✅ | ✅ |

The only built-in method that catches these invalid URLs is [decodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) though. That somewhat makes sense; it is the only method that actually needs to interpret the escape sequences and decode utf-8. It also seems behave consistently across all environments, including NodeJS.

Still, you will notice I opted to write our own solution. My main motivation is to not depend on a method that honestly is a bit of a black box, so it may end up truncating to (near) empty strings, if something else in the URL is invalid. The implementation proposed specifically checks for certain markers, so **it just leaves content it doesn't deem invalid as-is**. (Take some time to appreciate that double negative). I personally think this is better, because this is client-side. I'd rather see the malformed URLs (not caused by this truncation) come in server-side, if there is another error than incomplete escape sequences in there.

Of course, you may still voice your opinion if you prefer the `decodeURIComponent` implementation:
```typescript

function isValidUrl(input: string): boolean {
    try {
        decodeURIComponent(input);
        return true;
    } catch (ignored) {
        return false;
    }
}

/**
 * Truncate a URL to an arbitrary length, taking care to not break inside a percent escape or UTF-8 multibyte sequence.
 *
 * @param input The input to truncate to the specified limit.
 * @param limit The limit to apply; if negative, no truncation is applied.
 * @returns The URL, possibly truncated to a length near limit (at most 11 characters less than limit).
 */
export function truncateUrl(input: string, limit: number): string {
    if (limit < 0 || input.length <= limit) {
        return input;
    }
    let truncated = input.substring(0, limit);
    while (truncated.length > 0 && !isValidUrl(truncated)) {
        truncated = truncated.substring(0, truncated.length - 1);
    }
    return truncated;
}
```

### Changing the limit

As a last side-note: I don't know why the author of #405 decided to go with 128 as limit for some URLs. It seems our schema states 1024 as limit for those URLs. I don't mind increasing those truncation limits, but I believe that belongs in a separate PR from this issue fix.

### How to test

I wrote unit tests that test input strings on multiple markers, with many scenarios. When switching to the alternative implementation, every test except `URL_WITH_INVALID_ESCAPES` behaves the same. (The latter for obvious reasons).

### Checklist:

-   [X] Unit tests and/or functional tests are written and cover the proposed changes :exclamation:
-   [X] The proposed change can't affect any current customer implementation that could lead to events being rejected from our APIs or events to be miscategorized by UA.

[UA-9204]: https://coveord.atlassian.net/browse/UA-9204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/utils.ts
+++ b/src/client/utils.ts
@@ -15,3 +15,68 @@ export function isObject(o: any): boolean {
 export function coerceToNumber(input: any): any {
     return typeof input === 'string' && input != '' && !Number.isNaN(+input) ? +input : input;
 }
+
+/**
+ * Can be used to both check if the first bit is set (for any utf-8 multibyte part),
+ * and to detect "following bytes" (which start with `10xx_xxxx`)
+ */
+const UTF8_HIGH_BIT = 0b1000_0000;
+/** Header for a 2-byte code point: `110x_xxxx`. Can also be used as bit-mask to check for "following bytes". */
+const UTF8_HEADER_2 = 0b1100_0000;
+/** Header for a 3-byte code point: `1110_xxxx`. Can also be used as bit-mask to check for "header 2". */
+const UTF8_HEADER_3 = 0b1110_0000;
+/** Header for a 4-byte code point: `1111_0xxx`. Can also be used as bit-mask to check for "header 3". */
+const UTF8_HEADER_4 = 0b1111_0000;
+
+function utf8ByteCountFromFirstByte(firstByte: number): number {
+    if ((firstByte & 0b1111_1000) === UTF8_HEADER_4) {
+        return 4;
+    }
+    if ((firstByte & UTF8_HEADER_4) === UTF8_HEADER_3) {
+        return 3;
+    }
+    if ((firstByte & UTF8_HEADER_3) === UTF8_HEADER_2) {
+        return 2;
+    }
+    return 1;
+}
+
+/**
+ * Truncate a URL to an arbitrary length, taking care to not break inside a percent escape or UTF-8 multibyte sequence.
+ *
+ * @param input The input to truncate to the specified limit.
+ * @param limit The limit to apply; if negative, no truncation is applied.
+ * @returns The URL, possibly truncated to a length near limit (at most 11 characters less than limit).
+ */
+export function truncateUrl(input: string, limit: number): string {
+    if (limit < 0 || input.length <= limit) {
+        return input;
+    }
+    // A valid escape sequence is a percent followed by 2 hexadecimal characters; check if we split one up.
+    let end = input.indexOf('%', limit - 2);
+    if (end < 0 || end > limit) {
+        end = limit;
+    } else {
+        limit = end;
+    }
+    // Check that truncating at end won't break up an UTF-8 multibyte sequence half-way,
+    // by peeking backwards to find the first byte of an UTF-8 sequence (if present).
+    while (end > 2 && input.charAt(end - 3) == '%') {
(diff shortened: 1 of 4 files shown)


src/client/utils.ts
```diff
@@ -15,3 +15,68 @@ export function isObject(o: any): boolean {
...
+ * @param limit The limit to apply; if negative, no truncation is applied.
+ * @returns The URL, possibly truncated to a length near limit (at most 11 characters less than limit).
+ */
+export function truncateUrl(input: string, limit: number): string {
+    if (limit < 0 || input.length <= limit) {
+        return input;
+    }
+    // A valid escape sequence is a percent followed by 2 hexadecimal characters; check if we split one up.
+    let end = input.indexOf('%', limit - 2);
+    if (end < 0 || end > limit) {
```
fvanham-coveo, line 57: how is this last disjunction possible?

fvanham-coveo (approved):
The 128 limit should read 1024 based on json schemas and table schemas. Should be a separate change.

src/client/utils.ts
rphilippen: I opted to not do an intermediate substring, rather to find the right truncation point and substring only once. That does mean that this `indexOf('%')`, which starts two characters before the limit, searches the original input (also beyond `limit`). That means it might find a `%` at a later index, which should be considered equal to not finding it.

aperron-coveo (approved):

rphilippen:
I'll merge this on Monday, and hopefully @fvanham-coveo can help me validate it works well after deployment.

PR #472, opened 2024-08-15
feat(SFINT-5673): UA event logTriggerNotify added to insight client

SimonMilord:
### Proposed changes:

The following custom event has been added to Insight Client for notify triggers in insight usecase:

- logTriggerNotify

[SFINT-5673]: https://coveord.atlassian.net/browse/SFINT-5673?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/insight/insightClient.ts
+++ b/src/insight/insightClient.ts
@@ -19,6 +19,7 @@ import {
     SmartSnippetFeedbackReason,
     SmartSnippetLinkMeta,
     SmartSnippetSuggestionMeta,
+    TriggerNotifyMetadata,
 } from '../searchPage/searchPageEvents';
 import {
     ExpandToFullUIMetadata,
@@ -685,6 +686,10 @@ export class CoveoInsightClient {
             SearchPageEvents.showLessFoldedResults,
             metadata ? generateMetadataToSend(metadata, false) : undefined
         );
+    }
+
+    public logTriggerNotify(metadata: TriggerNotifyMetadata) {
+        return this.logCustomEvent(SearchPageEvents.triggerNotify, metadata);
     }
 
     private async getBaseCustomEventRequest(metadata?: Record<string, any>) {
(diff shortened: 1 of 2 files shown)


src/insight/insightClient.ts
```diff
@@ -687,6 +688,10 @@ export class CoveoInsightClient {
         );
     }
 
+    public logTriggerNotify(metadata: TriggerNotifyMetadata) {
```
mmitiche, line 691: We usually include the case metadata in the custom events logged for insight panel, here is an example: https://github.com/coveo/coveo.analytics.js/blob/984b149ae27f548417dd5a9d7153c7f1e188d703/src/insight/insightClient.ts#L595-L605

SimonMilord: oh right, will add !

mmitiche (approved):
LGTM!

erocheleau (approved):
Is there an EP version of this?

jstpierref:
> Is there an EP version of this?

@erocheleau No!

lbourdages (commented):
Overall it looks okay but I want to challenge the addition of new stuff to this library which we're trying to get rid of. Shouldn't we at least make that available in EP as Étienne suggested?

erocheleau:
> Overall it looks okay but I want to challenge the addition of new stuff to this library which we're trying to get rid of. Shouldn't we at least make that available in EP as Étienne suggested?

This already exists in UA, we're just adding it to our Insight version of the UA library because it was missing.

For EP we're still trying to figure out if we could perhaps rely on a server-side event instead of a client-side that's why we're not adding it to anything EP for now.

And we won't get rid of this library any time soon anyway :)

lbourdages (approved):

PR #205, opened 2020-06-18
Add analytics support for Pager/Facet/Query error in SearchPageClient

olamothe:
Nothing super fancy ;)

PR #488, opened 2025-09-01
chore: fix build failing on assert keyword

rphilippen:
### Proposed changes:

Our infra has a pending update from NodeJS version 20 → 22 that fails, because of the `assert` keyword found in the rollup config. This keyword was part of an early version/iteration of the [Import Attributes specification, but has been replaced with `with`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with). Support for `assert` was removed from NodeJS, so you now get this error:
```
> coveo.analytics@2.30.45 build
> rollup -c

[!] SyntaxError: Unexpected identifier 'assert'
    at compileSourceTextModule (node:internal/modules/esm/utils:346:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:107:18)
    at ModuleLoader.#translate (node:internal/modules/esm/loader:540:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:587:27)
    at ModuleJob._link (node:internal/modules/esm/module_job:162:19)
```

The `prettier` version used did not recognize the `with` syntax properly, so I had to update that too. While I was at it, I did an `npm audit fix` to ensure some `devDependencies` are using newer versions with less known vulnerabilities.

I ran `npm lint:fix` and it changed a lot of files. I put these changes in a separate commit so you can review them separately. I'm not sure if this is expected, or if we need to tweak something in `.prettierrc` and/or `.prettierignore`.

⚠️ Then I ran into what I think is a mistake, which was made 5 years ago in PR #205. I had to comment out a mapping to make the build work again, it seems TypeScript has gotten stricter over the years, and nobody ran the tests again? It is rightfully complaining the same constant is being defined twice in the mapping. I will add a comment on the relevant place, as it is easier to discuss the problem and fix in a comment thread.

### How to test

- Ensure you're on NodeJS version ≥ 22.
- Pull in the branch
- Run `npm install` followed by `npm run build`.
- For good measure, run `npm run test`.

The Rollup build completes successfully, instead of giving the error mentioned above. The tests pass.

### Checklist:

-   [X] ~Unit tests and/or functional tests are written and cover the proposed changes :exclamation:~ _Not applicable._
-   [X] The proposed change can't affect any current customer implementation that could lead to events being rejected from our APIs or events to be miscategorized by UA.

[UA-4616]: https://coveord.atlassian.net/browse/UA-4616?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aperron-coveo (approved):

src/version.ts
```diff
@@ -1 +1 @@
-export const libVersion = process.env.PKG_VERSION || 'local'; // processed by @rollup/plugin-replace
+export const libVersion = 'local'; // processed by @rollup/plugin-replace
```
tsteijvers, line 1: why was this removed? (could not find a reference in the description)

src/searchPage/searchPageEvents.ts
```diff
@@ -361,7 +361,8 @@ export const CustomEventsTypes: Partial<Record<SearchPageEvents | InsightEvents,
     [SearchPageEvents.triggerExecute]: 'queryPipelineTriggers',
     [SearchPageEvents.triggerQuery]: 'queryPipelineTriggers',
     [SearchPageEvents.triggerRedirect]: 'queryPipelineTriggers',
-    [SearchPageEvents.queryError]: 'errors',
+    // This value is the same as SearchPageEvents.triggerQuery, which is very likely a mistake.
```
tsteijvers, line 364: I must be misreading the code...  the value of `SearchPageEvents.triggerQuery` is "queryPipelineTriggers"...?

src/version.ts
rphilippen: The newer tooling complained that `'<replaced version>' || 'local'` contained dead code, so instead of using this I just replace the `'local'` string and there's no useless fallback.

rphilippen:
Apologies for the force push. I used `pnpm` instead of `npm`, and added its lockfile in the first commit 🤦‍♂️. This has been corrected.

Sadly, that does mean my vulnerability fix of mock-test doesn't work.

tsteijvers (approved):

fvanham-coveo (approved):
Looks ok, but why is the formatter injecting `,` after the last parameter in a function call?

olamothe (approved):

rphilippen:
> Looks ok, but why is the formatter injecting `,` after the last parameter in a function call?

I think it's the same principle as the `,` on the last entry of an object too: laziness. Just put a comma after everything, so you don't have to think about whether it's the last entry 🤷‍♂️.

image

PR #500, opened 2026-05-14
feat(SFINT-6765): logGeneratedAnswerFollowupOpenSource method created…

erocheleau:
### Proposed changes:

- Created the methods logGeneratedAnswerFollowupOpenSource and makeGeneratedAnswerFollowupOpenSource, these new events will be triggered whenever a user clicks on the citation offered by the Search Agent on a follow-up answer only.

***Not to be confused with the separate event we use for a click on the citations on the first answer generated***

The reason behind can be viewed as part of this RFC: https://coveord.atlassian.net/wiki/spaces/SFINT/pages/6382813250/Decision+on+tracking+clicks+on+Citations+in+follow-up+answers

TL;DR We don't want user clicks on follow-up citations to be UA click events since it doesn't make sense for those to be clicks associated with the original search that started the conversation. So we made those custom events instead.

### How to test

`npm run test`

### Checklist:

- [X] Unit tests and/or functional tests are written and cover the proposed changes :exclamation:
- [X] The proposed change can't affect any current customer implementation that could lead to events being rejected from our APIs or events to be miscategorized by UA.

[SFINT-6765]: https://coveord.atlassian.net/browse/SFINT-6765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lbourdages (approved):

src/searchPage/searchPageEvents.ts
```diff
@@ -358,6 +358,10 @@ export enum SearchPageEvents {
      * Identifies the click event that gets logged when the user clicks on a generated answer citation.
      */
     generatedAnswerCitationClick = 'generatedAnswerCitationClick',
+    /**
+     * Identifies the click event that gets logged when the user clicks on a generated answer citation.
```
mmitiche:
```suggestion
     * Identifies the custom event that gets logged when the user clicks on a generated answer citation.
```

mmitiche (approved):
LGTM!

src/searchPage/searchPageEvents.ts
erocheleau: Lol that comes back to bite me hahaha

felixpelletier (approved):

mmitiche (approved):

felixpelletier (approved):
turbo approved

mmitiche (approved):

